PR #158, opened 2014-03-11
Use Node Metadata to Over Ride Node Attributes

ChrisPortman:
Where appropriate, allow node metadata to take precedence over node attributes applied by a policy in places such as Tasks and Brokers.

Just looking for feedback atm, I haven't had a chance to run the tests on this yet

### DIFF
--- a/brokers/chef.broker/install.erb
+++ b/brokers/chef.broker/install.erb
@@ -56,7 +56,7 @@ printf '%s\n' "$rsa_key" >> /etc/chef/validation.pem
 cat <<'EOP'
 	log_level :info
 	log_location STDOUT
-	chef_server_url "<%= broker.chef_server_url %>"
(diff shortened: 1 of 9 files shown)


smcclellan:
I'm +1 on pulling this in. We need to:
1. Rebase it
2. Document in each task/broker's README.md which node metadata will be used
3. Be consistent about whether these metadata are in the core utilities or in task/broker code. I lean toward the latter, which will make future extensibility easier (and prevent unexpected effects/quirks).

Most tasks have already taken this approach with things like the hostname, and I think as long as we keep the onus on customization files (task/broker/hook), we'll end up with more portable and robust stock offerings.

smcclellan:
This is already adopted as a best practice (thanks to your work, primarily). The code in this PR is mostly already included. I'm not sure if modifying the helper functions is a good practice at this point due to potential backwards-compatibility breakage. So with that said, I'm closing this, and any missing metadata checks can be modified as a separate PR. Thanks!

PR #174, opened 2014-04-03
(RAZOR-190) Add command to update repo

ChrisPortman:
Add an update-repo command that will allow the update of repo settings. When the supplied setting are different, the process will trigger a 'make_repo_available' event.  This is required, because if you create the repo and then policies/tags etc only to realise you botched the definition of the repo, you first have to pull down all the dependants (policies and the like) to delete/recreate the repo.

### DIFF
--- a/lib/razor/command/refresh_repo.rb
+++ b/lib/razor/command/refresh_repo.rb
@@ -1,0 +1,43 @@
+# -*- encoding: utf-8 -*-
+class Razor::Command::RefreshRepo < Razor::Command
+  summary "Initiate a refresh of a repository from its source"
+  description <<-EOT
+Initiates a refresh of a repoisitory from its source.  This is done by removing
+the repository from disk and replacing it with a fresh copy from the source.
+
+This is also useful in cases where multiple razor servers are connected to
+a single database.  Second and subsequent servers can use this command to
+populate their own copies of repositories without having to define their own
+in the database.
+
+WARNING: This operation will cause an interuption in the repositories availability
+and may impact dependant tasks in progress.
+  EOT
+
+  example <<-EOT
+Refresh an existing repository:
+  {
+    repo: { 'name': 'myRepo' }
+  }
+  EOT
+
+  authz '%{repo}'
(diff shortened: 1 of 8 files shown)


lib/razor/command/refresh_repo.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def run(request, data)
+    repo = Razor::Data::Repo[:name => data['repo']['name']]
+    repo.refresh(@command)
+    { :result => "repo refresh started" }
+  end
+
+  def self.conform!(data)
+    data.tap do |_|
+      data['repo'] = { 'name' => data['repo'] } if data['repo'].is_a?(String)
```
smcclellan: This `conform!` method is primarily to facilitate backwards compatibility for commands. Our other APIs have moved in favor of just accepting the name of a datatype, so I'd stick to that for new endpoints. If a case ever comes up where we need to also accept an object like this, we can add the code back in.

lib/razor/data/repo.rb
```diff
@@ -170,7 +190,8 @@ def filesystem_safe_name
     # done, notify ourselves of that so any cleanup required can be performed.
     def unpack_repo(command, path)
       destination = iso_location
-      destination.mkpath        # in case it didn't already exist
+      FileUtils.remove_entry_secure(destination) if destination.directory? #delete the destination and contents if it already exists
```
smcclellan: This has failed in the past if any files in the repo are not extracted with write permissions. The fix, already implemented, is to recursively set this privilege before attempting the deletion.

smcclellan:
I think the cleaner approach to this problem will be to add an `update-policy-repo` command. With that, a user can solve re-downloading/re-unpacking an ISO while maintaining live install integrity through some immutability at the repo level.

That command would get you the ability to salvage valid policies when the repo's source is incorrect due to typo or not-ready-yet 404.

smcclellan:
Closing due to inactivity. The ticket to add an `update-policy-repo` command is linked above. A command like that should resolve the use case(s) in this ticket.

PR #186, opened 2014-04-15
Add like operator to Matcher

LordGaav:
I added an extra operator called 'like' to the Matcher, so you can use Regex matching. In our use case, we use a Regex to match a certain range of mac addresses.


smcclellan:
Tracked instead by https://github.com/puppetlabs/razor-server/pull/287.

PR #193, opened 2014-04-23
Cancel commands

smcclellan:
Currently, if the user mistypes the URL for the ISO in a repo, that command will continue retrying. This command allows the user to transition a command instance into a new 'cancelled' status so it will not continue processing.


smcclellan:
Let's keep repos a bit more immutable. If a URL is typed wrong, delete and recreate the repo. Error messages should fall in line.

See [this ticket](https://tickets.puppetlabs.com/browse/RAZOR-762) for the work on making `create-repo` commands stop querying an ISO URL when the repo itself is deleted.

PR #204, opened 2014-05-12
Added detection of Windows Kits Folder

wcbzero:
Added detection of Windows Kits Folder from the Windows Registry, and a functionality to import drivers into the wim, by putting the desired INF files in $cwd\drivers.


smcclellan:
Functionality duplicated by [other initiatives](https://github.com/puppetlabs/razor-server/pull/319).

PR #253, opened 2014-09-18
(RAZOR-380) Normalize inconsistencies in attribute names

justinclayton:
These are the server-side fixes for [RAZOR-380](https://tickets.puppetlabs.com/browse/RAZOR-380)


smcclellan:
Thanks for the pull request! These changes have already been made. Our standard is that properties in the CLI will use a dash (e.g. `--root-password`), elsewhere (e.g. in JSON) will be handled as an underscore. Wherever possible in commands, we automatically accept both forms.

PR #261, opened 2014-12-11
(RAZOR-421) Support overriding protect_new_nodes by subnet

ChrisPortman:
Adding a configuration option called 'invert_protect_new_nodes_for_subnets' that accepts a list of subnets in CIDR notation (a.a.a.a/bb) for which the 'protect_new_nodes' should be reversed.  This will allow the user to set a default protection value and then override it for specific subnets.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -199,7 +199,7 @@ and requires full control over the database (eg: add and remove tables):
         net_id = "net#{index - 1}"
         vars[net_id] = "${#{net_id}/mac:hexhyp}"
       end
-      ["dhcp_mac", "serial", "asset", "uuid"].each { |k| vars[k] = "${#{k}}" }
+      ["dhcp_mac", "serial", "asset", "uuid", "ip_addr"].each { |k| vars[k] = "${#{k}}" }
       q = vars.map { |k,v| "#{k}=#{v}" }.join("&")
       url "/svc/boot?#{q}"
     end
--- a/config.yaml.sample
+++ b/config.yaml.sample
@@ -50,6 +50,12 @@ all:
 
   # Should newly discovered nodes be marked installed?
   protect_new_nodes: false
+  
+  # Invert the 'protect_new_nodes' value for the following subnets
+  # List subnets in the CIDR format (a.a.a.a/bb)
+  #
+  # invert_protect_new_nodes_for_subnets:
+  #   - 192.168.1.0/24
 
   # How to match nodes to possibly existing nodes in the database. The node
   # sends us the MAC addresses of its network interfaces, serial number,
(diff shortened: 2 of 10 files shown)


lib/razor/data/node.rb
```diff
@@ -181,6 +183,26 @@ def registered?
       ! facts.nil?
     end
 
+    #Determine if this machine should be protected as a new machine based
+    #on the value of the config param 'protect_new_nodes' and 'invert_protect_new_nodes_for_subnets'
+    def protect?
+      protected = Razor.config['protect_new_nodes'] || false
+      override_subnets = Razor.config['invert_protect_new_nodes_for_subnets'] || []
+
+      if self.metadata['ip']
```
smcclellan: Is this metadata value set so consistently that it can reliably factor into deciding whether to provision a node? Where is this {{ip}} node metadata value coming from, if not the task's {{store_ip}} function when the node has just completed installation? I can't see a picture here of how this works when {{protect_new_nodes}} is false.

config.yaml.sample
```diff
@@ -50,6 +50,12 @@ all:
 
   # Should newly discovered nodes be marked installed?
   protect_new_nodes: false
+  
+  # Invert the 'protect_new_nodes' value for the following subnets
+  # List subnets in the CIDR format (a.a.a.a/bb)
+  #
+  # invert_protect_new_nodes_for_subnets:
```
smcclellan: I'm wary of a property that inverts the value of another property. Toggling `protect_new_nodes` completely changes the behavior of this property. One could feasibly toggle the above value without reading this comment. Did you consider a less contingent property or properties, maybe `protect_new_nodes_on_subnet` and/or `provision_new_nodes_on_subnet`?

app.rb
```diff
@@ -199,7 +199,7 @@ def ipxe_boot_url
         net_id = "net#{index - 1}"
         vars[net_id] = "${#{net_id}/mac:hexhyp}"
       end
-      ["dhcp_mac", "serial", "asset", "uuid"].each { |k| vars[k] = "${#{k}}" }
+      ["dhcp_mac", "serial", "asset", "uuid", "ip_addr"].each { |k| vars[k] = "${#{k}}" }
```
smcclellan: Is there a reason not to include this, regardless of the subnet functionality included in this PR? I'm not sure if this value is absent in any cases, or why it was omitted originally.

smcclellan:
Closing due to inactivity. Feel free to open a new PR if the issue persists. A better solution may be to support `protect_new_nodes` being a nested object rather than just the current boolean, enabling ease of backwards compatibility as well as one contained property for how `protect_new_nodes` works.

PR #262, opened 2014-12-11
add support for xenserver 6.2

kongluoxing:
add task scripts for xenserver, tested on xenserver 6.2

### DIFF
--- a/tasks/xenserver.task/README.md
+++ b/tasks/xenserver.task/README.md
@@ -1,0 +1,7 @@
+# Installation Tips
+Befor installation please make sure these preparation step is done:
+```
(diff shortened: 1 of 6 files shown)


smcclellan:
Per the above comment, it would be best if we can avoid the manual `zcat` step in one of our stock tasks. In order for us to properly add a Xen task, I have created two tickets to make this easier for users:
- https://tickets.puppetlabs.com/browse/RAZOR-930 to allow a `post_extract_script` value in `metadata.yaml`
- https://tickets.puppetlabs.com/browse/RAZOR-931 to track adding a Xen stock task.

Thanks for the contribution! I am closing this PR in the meantime.

PR #265, opened 2014-12-26
(RAZOR-433) Facilitate the retrival of full objects in collection

ChrisPortman:
Add an "expand" parameter to collections so that the full opbjects are listed in a collection. E.g. /api/collections/nodes?expand=true will return all nodes with all their details.

This will better support interface integration to the API where previously, one would have to issue a call against the nodes collection and then another call for each of the nodes returned to build the full picture.  This allows the same in one call.


smcclellan:
Thanks for the PR! We've merged functionality using the `depth` argument here: https://github.com/puppetlabs/razor-server/pull/405.

PR #267, opened 2015-01-05
(RAZOR-436) build razor-server for el7

melissaanne:
This commits allows razor-server packages to be built for el7


melissaanne:
In order to install this on el7, we first need to build and ship razor-torquebox

smcclellan:
Rebased, updated commit message, and merged as c4dc720.

PR #268, opened 2015-01-06
(RAZOR-438) Do not build razor-server for el5

melissaanne:
Razor server has a dependency on `libarchive-devel`, which is not
available on el5. Because of this, we will not be providing razor server
packages for el5 until this dependency requirement is met.

Addresses https://tickets.puppetlabs.com/browse/RAZOR-438


melissaanne:
Turns out, I was too hasty, and forgot about the EPEL repo. The EPEL repo does provide libarchive, so we are good to include el5 in our release!

PR #269, opened 2015-01-08
Fixing entry serialization

smcclellan:
The same serialization issue that has affected policies, hooks, etc. also
affected events. The validation plugin was validating an object before
the serialization plugin acts on the object to make it valid.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-439


lutter:
ACK

PR #270, opened 2015-01-14
Report "stage-done" for brokers

smcclellan:
This changes the stock brokers to cause a note to go
into the node's log that the broker has finished and
the node is transitioned out of Razor.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-357


holguinj:
:+1: from me after walking through the changes with @smcclellan.

PR #272, opened 2015-01-19
Fix psmodule path to find dism

nickperry:


nickperry:
Accidentally also included ba4fd15 in this too which I just sumbitted separately in as pull request #271

PR #274, opened 2015-01-19
fix subtle weird hyphen ultimate time sink problem

nickperry:
note the corrected hyphen before PassThru. It's subtle but it was erroneously an en dash instead of a hyphen (you can see it was ever so slightly longer). It was breaking the script.


smcclellan:
Rebased, revised commit message, and merged.

PR #275, opened 2015-01-23
Run Windows broker in SetupComplete.cmd

smcclellan:
The previous method of creating a service was not saving
the broker script in an accessible location. This method
instead stores a script in the SetupComplete.cmd file that
will do both the download and the execution.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-471

### DIFF
--- a/tasks/windows.task/SetupComplete.cmd.erb
+++ b/tasks/windows.task/SetupComplete.cmd.erb
@@ -1,0 +1,8 @@
+# This file will be executed after reboot but before the login screen
+# is displayed.
+# The script itself does several things:
+# - Gets the proper temporary folder to store the broker file.
+# - Downloads the broker to this folder.
+# - Runs the broker.
+# - Deletes the file it downloaded.
+powershell.exe -executionpolicy bypass -command "set-variable -name temp -value (join-path -path (Get-ItemProperty -Path 'HKLM:\System\CurrentControlSet\Control\Session Manager\Environment' -Name TEMP).TEMP -childpath razor-broker.ps1); (New-Object System.Net.WebClient).DownloadFile('<%= broker_install_url('install.ps1') %>',$temp);invoke-expression $temp;remove-item $temp"
--- a/tasks/windows.task/second-stage.ps1.erb
+++ b/tasks/windows.task/second-stage.ps1.erb
@@ -29,21 +29,10 @@ start-process "${installer}" -argumentlist "/unattend:${unattended} /noreboot" -
 write-host "notify Razor that the installer completed"
 (new-object System.Net.WebClient).DownloadString('<%= stage_done_url("finished") %>')
 
-write-host "download broker script"
-$broker_url = "<%= broker_install_url('install.ps1') %>"
-# This file needs to have a .ps1 extension for when it's passed to Powershell.
-$broker_path = [IO.Path]::GetTempFileName() | Rename-Item -NewName { $_ -replace 'tmp$', 'ps1' } -PassThru
-(new-object System.Net.WebClient).DownloadFile($broker_url, $broker_path)
-
-write-host "set broker to run after reboot"
-$powershell_path = $(get-process powershell | select -first 1).path
-# When this service runs, it results in a "The service did not respond to
-# the start or control request in a timely fashion" error message. This message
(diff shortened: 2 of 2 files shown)


nickperry:
Unfortunately this won't work either. You're adding your key to RunOnce in the WinPE environment's registry. I think we need to use the reg command to load the hive from the newly installed environment under D:

nickperry:
The following works:

reg load HKLM\SOFTWARE2 D:\Windows\System32\config\SOFTWARE
set-location HKLM:\SOFTWARE2\Microsoft\Windows\CurrentVersion\Run

new-itemproperty . Razor -propertytype String -value "Powershell -executionpolicy bypass -command C:\Windows\Temp\razor-broker.ps1"

reg unload HKLM\SOFTWARE2

smcclellan:
Excellent, thanks for that! I was just looking into adding the registry key in the D: drive's registry after realizing that winpe has its own registry. I'll test this out and update this PR as needed.

nickperry:
The next problem though is that the broker script will only get invoked via RunOnce the first time somebody interactively logs into the machine.

I guess the ultimate solution is to use reg load as I mentioned above to offline load the SYSTEM hive, and instead of adding a key to RunOnce, add a service under HKLM\\SYSTEM\CurrentControlSet\Services

smcclellan:
Yes, that's a combination of HEAD's current solution and this PR's solution. I will try that out.

nickperry:
Great. I guess either do it all manually with New-Item to create the key and then a few New-ItemPropertys or you could define the service on your dev machine, export its key to a .reg file and get second-stage.ps1 to write out something similar (with an updated path) which can then be imported into the offline SYSTEM hive using reg import.

smcclellan:
@nickperry — Do you know whether it's possible to download miscellaneous files via `initrd` that can be made available inside WinPE? Something like:
```
initrd ${base}/test.txt                     X:\test.txt # or D:\test.txt?
```
This might not be used for this particular issue, but could help for extracting the razorurl argument from the build-razor-winpe.ps1 script, making razor's winpe more generic.

nickperry:
@smcclellan So I had a play about with injecting arbitrary files through initrd lines but it doesn't seem to work I'm afraid. Shame, as it would be great to save people all the hassle of fiddling about with ADK. That said, I would think it's highly likely that licensing constraints would make it impossible to distribute a generic razor winpe image.

smcclellan:
@nickperry — I figured that might be the case with initrd.

As for this PR, it looks like Windows requires more than just the registry item in `HKLM:\SYSTEM\CurrentControlSet\Services` to install a service. I exported an existing service entry, modified the service name, and reimported, but that did not place the new service in the Services window. Is there another behind-the-scenes Powershell command for creating services this way?

This might be relevant: "You must use the functions provided by the SCM, instead of modifying the database directly." from https://msdn.microsoft.com/en-us/library/windows/desktop/ms682544%28v=vs.85%29.aspx.

nickperry:
@smcclellan :( You're right, my reg suggestion doesn't work. This is quite frustrating. I just can't seem to find a way to achieve this.

smcclellan:
@nickperry — I found a direction that will work. This PR now installs a SetupComplete.cmd file that gets executed at the right time: After reboot, before the login screen is displayed. I have tested that this script is called successfully in my infrastructure.

nickperry:
@smcclellan Excellent. I'll give it a go later today. I looked at setupocomplete.cmd a couple of weeks ago, but didn't end up pursuing it. Really happy to hear it works.

nickperry:
@smcclellan Works perfectly for me. Great to get this bit sorted finally.

nickperry:
@smcclellan I've been testing a little more and have observed sometimes my new windows installation is mounted under E:\ rather than D:\. I've not as yet found the "proper" way to detect this. But we do need a way to handle it.

smcclellan:
@nickperry I have created https://tickets.puppetlabs.com/browse/RAZOR-484 to track the generalization of the drive letter. I have read up more and apparently this should be done via BCD, but I haven't found the specific incantation to query that data yet. BCD would be preferred to querying each drive letter for some representative file. If you find any more on how to determine this, please post in that ticket.

PR #276, opened 2015-01-29
Syncing the time before we communicate with puppet master in case the ha...

tcf909:
...rdware clock is wrong or the hwclock  was set wrong because of utc issues (virtual environment).

### DIFF
--- a/brokers/puppet.broker/install.erb
+++ b/brokers/puppet.broker/install.erb
@@ -36,6 +36,9 @@ else
     fail "neither yum or apt are installed, so I can't figure out what next!"
 fi
 
+#sync the time if possible
+yum -y install ntpdate && ntpdate us.pool.ntp.org
+
 
 # Now we have that, we can install our platform specific package to define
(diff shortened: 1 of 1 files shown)


brokers/puppet.broker/install.erb
```diff
@@ -36,6 +36,9 @@ else
     fail "neither yum or apt are installed, so I can't figure out what next!"
 fi
 
+#sync the time if possible
+yum -y install ntpdate && ntpdate us.pool.ntp.org
```
smcclellan: It would help to allow apt-get for Debian systems too.

smcclellan:
This issue is more pervasive than just the puppet broker. Not all installs will need this, I think. I have created [a ticket](https://tickets.puppetlabs.com/browse/RAZOR-933) to track progress on this issue and ensure we test and document the change. I will close this PR now in favor of a more generic fix in https://github.com/puppetlabs/razor-server/pull/355.

PR #277, opened 2015-02-17
Securing /api endpoint with `secure_api` config

smcclellan:
The /api namespace would benefit from security control. The actual configuring
of ports and certificates is done in Torquebox configuration, but we can ensure
that communications are secure in each call to /api endpoints. This change adds
a config property called `secure_api` that, when enabled, returns a 401
(Unauthorized) when an /api endpoint is called without HTTPS. By default,
this is disabled (same as current behavior).

Fixes https://tickets.puppetlabs.com/browse/RAZOR-487

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # Razor Server Release Notes
 
+## Next - TBD
+
++ NEW: The `secure_api` config property can be used to ensure that
+  communications with /api are secure. When this is true, all calls to the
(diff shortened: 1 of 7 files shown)


holguinj:
:+1: from me

lutter:
Looks good to me - the only issue is that 401 is the wrong status code to return here. A 404 would be better, since there is nothing the user can do (like supply credentials) that would make the request succeed.

smcclellan:
Changed to 404 and rebased.

PR #279, opened 2015-02-25
(RAZOR-499) Support Structured Metadata

ChrisPortman:
This allows for structured data to be stored as values to metadata similar to structured facts.

The values can be specified as native json where raw json is being sent to the api, or as a single quoted string on the command line.


smcclellan:
Rebased and merged.

PR #280, opened 2015-03-03
Standardize API argument names from '-' to '_'

smcclellan:
In order to standardize the API more, we will use underscores in argument names
rather than dashes. The previous dashes will still be accepted as input, but
the returned values and documentation should be exclusively underscores.

Additionally, if both the dash and underscore arguments are supplied, an
error will be thrown.

For https://tickets.puppetlabs.com/browse/RAZOR-481

### DIFF
--- a/lib/razor/command/create_broker.rb
+++ b/lib/razor/command/create_broker.rb
@@ -16,7 +16,7 @@ Creating a simple Puppet broker:
          "server":      "puppet.example.org",
          "environment": "production"
       },
-      "broker-type": "puppet"
+      "broker_type": "puppet"
     }
   EOT
 
@@ -35,7 +35,7 @@ Creating a simple Puppet broker:
     which broker the node will be handed off via after installation.
   HELP
 
-  attr 'broker-type', required: true, type: String, references: [Razor::BrokerType, :name],
+  attr 'broker_type', required: true, type: String, references: [Razor::BrokerType, :name],
                       help: _(<<-HELP)
     The broker type from which this broker is created.  The available
     broker types on your server are:
@@ -44,7 +44,7 @@ Creating a simple Puppet broker:
 
   object 'configuration', help: _(<<-HELP) do
     The configuration for the broker.  The acceptable values here are
-    determined by the `broker-type` selected.  In general this has
+    determined by the `broker_type` selected.  In general this has
     settings like which server to contact, and other configuration
     related to handing on the newly installed system to the final
     configuration management system.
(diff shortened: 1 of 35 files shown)


lib/razor/command/create_broker.rb
```diff
@@ -24,7 +24,7 @@ class Razor::Command::CreateBroker < Razor::Command
 Creating a simple Puppet broker:
 
     razor create-broker --name puppet -c server=puppet.example.org \\
-        -c environment=production --broker-type puppet
+        -c environment=production --broker_type puppet
```
lutter: This is where underscores get kinda ugly: on the command line, words are usually separated by '-'. Since we will accept either, I'd prefer if the command line examples continue to use '-'

lib/razor/command/create_broker.rb
```diff
@@ -64,7 +63,8 @@ def run(request, data)
   def self.conform!(data)
     data.tap do |_|
       # Allow "c" as a shorthand.
-      add_hash_alias(data, 'configuration', 'c')
+      add_hash_alias(data, 'c', 'configuration')
+      add_alias(data, 'broker-type', 'broker_type')
```
lutter: Instead of requiring that command authors remember to set up aliases for the attributes that have '_' in them, is there a way we could make that automatic ? IOW, if you say "attr 'foo_bar'" that the machinery automatically sets up an alias for 'foo-bar' ?

lib/razor/command/create_broker.rb
smcclellan: That's fair; I'll update the examples accordingly.

lib/razor/command/create_policy.rb
```diff
@@ -133,9 +133,9 @@ def run(request, data)
 
     data["enabled"] = true if data["enabled"].nil?
 
-    data["max_count"] = data.delete("max-count") if data.has_key?("max-count")
-    data["root_password"] = data.delete("root-password") if data.has_key?("root-password")
-    data["node_metadata"] = data.delete("node-metadata") if data.has_key?("node-metadata")
+    data["max_count"] = data.delete("max_count") if data.has_key?("max_count")
```
lutter: These are now noops and can be deleted.

lutter: Also a noop

lib/razor/command/create_broker.rb
smcclellan: This would indeed be nice; working on it now.

smcclellan:
@lutter: I have revised the commit per your comments and added another to address the `alias` suggestion. I'm concerned there might be a name conflict down the road since `alias` is a reserved word in Ruby and we have `def alias(...)` to provide that functionality right now.

PR #282, opened 2015-03-04
Add `http_port` argument for bootstrap over HTTPS

smcclellan:
Since booted machines cannot by default use the HTTPS port, we need a way to
generate a bootstrap file that points to the correct HTTP port. That port is
managed outside of Razor, so that requires a new argument, `http_port`, to
the bootstrap endpoint that will be used in the generated bootstrap file.

For https://tickets.puppetlabs.com/browse/RAZOR-487

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -676,3 +676,12 @@ culminates in chain loading from the Razor server)
 The URL accepts the parameter `nic_max` which should be set to the maximum
 number of network interfaces that respond to DHCP on any given machine. It
 defaults to 4.
+
+If this request to generate the bootstrap file occurs over HTTPS, the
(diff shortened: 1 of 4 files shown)


doc/api.md
```diff
@@ -676,3 +676,7 @@ culminates in chain loading from the Razor server)
 The URL accepts the parameter `nic_max` which should be set to the maximum
 number of network interfaces that respond to DHCP on any given machine. It
 defaults to 4.
+
+If this request occurs over HTTPS, the `http_port` parameter must be supplied
```
lutter: Would probably be good to spell out what goes into the http_port. Right now, this is more a justification of why we need to know that - for the user, it's more important to know what information to supply.

lutter:
ACK. Looks good to me

PR #288, opened 2015-03-12
Rhel subscriptions

smcclellan:
Users will want RHEL systems to be subscribed to RHEL automatically. That can
be done in the kickstart file, but will require some credentials to perform.
There are two methods of doing this:
- Providing the activationkey
- Using auto-attach
Both require a username and password. So, the node metadata used for the
policy is now:
- `rhn_username`
- `rhn_password`
- (optional) `rhn_activationkey`
If the `rhn_activationkey` is omitted, this will use the auto-attach method.
If both `rhn_username` and `rhn_password` are omitted, this step will be
skipped.

The recommended way to add the metadata for this is through the policy's
`node_metadata` property.

This also adds timezone configuration.

For https://tickets.puppetlabs.com/browse/RAZOR-324


tasks/redhat.task/kickstart.erb
```diff
@@ -11,7 +11,20 @@ rootpw <%= node.root_password %>
 network --hostname <%= node.hostname %>
 firewall --enabled --ssh
 authconfig --enableshadow --passalgo=sha512 --enablefingerprint
-timezone --utc America/Los_Angeles
+timezone --utc <%= node.metadata['timezone'] || 'America/Los_Angeles' %>
+rhel_username="<%= node.metadata['rhn_username'] %>"
```
lutter: Lines 15 - 27 need to go into a %post script. See the [RHEL docs](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Installation_Guide/sect-kickstart-syntax.html#sect-kickstart-postinstall) for details on the syntax.

In a nutshell, only specific kickstart directives are legal in the main part of a kickstart file; to run a shell script, you need to enclose it in `%post ... %end`

lutter: Also, there's a command 'ksvalidator' which you can feed a kickstart file and have it check for syntax errors

lutter:
As for reorganizing stock tasks: how are users going to migrate their existing policies ? Even if we don't do it automatically, we need to at least explain to them the commands they need to run to do it.

What do you think of adding README.md's to each task that at least explain the node metadata that can be used to influence the task's behavior and what of that metadata must be set vs being optional ?

smcclellan:
I think the README idea is good. I'll add that to the existing stock tasks, mentioning which metadata are used.

tasks/redhat.task/README.md
```diff
@@ -0,0 +1,19 @@
+# Task notes for Redhat 6
```
lutter: This file seems to be a copy of `6.task/README.md`

lutter:
ACK. Looks good.

tasks/redhat.task/README.md
smcclellan: Yes; `redhat` is a shortcut for `redhat/6`. `centos` is also a shortcut for `centos/6`.

PR #290, opened 2015-03-14
Add config defaults override

smcclellan:
This creates an override capability for the config.yaml file. The config.yaml
must still exist, but if a property is absent from it, the defaults file will
be searched as well.

The RAZOR_CONFIG_DEFAULTS environment setting will go alongside RAZOR_CONFIG
for determining the location of these two files, the defaults and overrides
respectively.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-522


lutter:
ACK

smcclellan:
Rebased and merged.

PR #299, opened 2015-04-10
Guess HTTP port when generating bootstrap file

smcclellan:
The decision tree for what port to use for the bootstrap URL is now as follows:
- Is a valid `http_port` argument supplied? Use that. (Throw error on invalid.)
- Is the HTTP_PORT environment variable defined? Use that.
- Is the request secure? Use 8151 as a constant default.
- Otherwise, use the port where the request was received.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-549

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -780,11 +780,17 @@ and requires full control over the database (eg: add and remove tables):
       error 400,
         :error => _("The nic_max parameter must be an integer not starting with 0")
 
-    params['http_port'].nil? and request.secure? and
-      error 400,
-        :error => _("The `http_port` argument must be supplied for bootstrap generation on a secure port")
+    is_nil = params["http_port"].nil?
+    is_in_range = (params["http_port"] =~ /\A[1-9][0-9]{0,4}\Z/ and
+        params['http_port'].to_i < 65536)
+    is_nil or is_in_range or error 400,
+        :error => _('The http_port parameter must be an integer between 1 and 65535')
 
-    @bootstrap_port = params['http_port'].nil? ? request.port.to_s : params['http_port']
+    @bootstrap_port = params['http_port']
+    @bootstrap_port ||= ENV['HTTP_PORT']
+    # Can't use current port if this request is secure; use this default.
+    @bootstrap_port ||= 8150 if request.secure?
+    @bootstrap_port ||= request.port.to_s
 
     # How many NICs ipxe should probe for DHCP
     @nic_max = params["nic_max"].nil? ? 4 : params["nic_max"].to_i
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,9 @@
   addresses. 
 + NEW: The `str` tag matcher can be used to convert input (likely numeric)
   into a string.
++ IMPROVEMENT: The `bootstrap` URL will now guess what the correct http_port
+  value should be, typically falling back to the URL used for the `/bootstrap`
+  request. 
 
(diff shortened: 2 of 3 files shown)


app.rb
```diff
@@ -779,11 +779,15 @@ def render_template(name)
...
-      error 400,
-        :error => _("The `http_port` argument must be supplied for bootstrap generation on a secure port")
-
-    @bootstrap_port = params['http_port'].nil? ? request.port.to_s : params['http_port']
+    params["http_port"].nil? or (params["http_port"] =~ /\A[1-9][0-9]?[0-9]?[0-9]?[0-9]?\Z/ and
+      params['http_port'].to_i < 65536) or error 400,
+        :error => _('The http_port parameter must be an integer between 1 and 65535')
+
+    @bootstrap_port = params['http_port']
+    @bootstrap_port ||= ENV['HTTP_PORT']
```
lutter: How can this ever come into play ? We only get here if `params['http_port']` is a number so we would never look at the env variable.

smcclellan: This will be reached if the `http_port` parameter is not supplied. The `params["http_port"].nil?` short-circuits the rest of the part where we're checking that the parameter is an integer.

app.rb
```diff
@@ -779,11 +779,15 @@ def render_template(name)
       error 400,
         :error => _("The nic_max parameter must be an integer not starting with 0")
 
-    params['http_port'].nil? and request.secure? and
-      error 400,
-        :error => _("The `http_port` argument must be supplied for bootstrap generation on a secure port")
-
-    @bootstrap_port = params['http_port'].nil? ? request.port.to_s : params['http_port']
+    params["http_port"].nil? or (params["http_port"] =~ /\A[1-9][0-9]?[0-9]?[0-9]?[0-9]?\Z/ and
```
nathanielksmith: I have to google to find out what operator precedence in ruby with "and" / "or" is, which is a big hit to readability. perhaps clean up with parens or intermediate variables? this is really hard to read and parse.

smcclellan: The `and` is in parens. I agree that it could be made more readable though; let me see about refactoring a bit.

app.rb
nathanielksmith: i don't think `and` and `or` short circuit? isn't it best to use && and ||?

app.rb
```diff
@@ -779,11 +779,17 @@ def render_template(name)
       error 400,
         :error => _("The nic_max parameter must be an integer not starting with 0")
 
-    params['http_port'].nil? and request.secure? and
-      error 400,
-        :error => _("The `http_port` argument must be supplied for bootstrap generation on a secure port")
-
-    @bootstrap_port = params['http_port'].nil? ? request.port.to_s : params['http_port']
+    is_nil = params["http_port"].nil?
+    is_in_range = (params["http_port"] =~ /\A[1-9][0-9]?[0-9]?[0-9]?[0-9]?\Z/ and
```
pcarlisle: Seems a little more readable as
`/\A[1-9][0-9]{4}\Z/`

aperiodic: Wouldn't it be `/\A[1-9][0-9]{0,4}\Z/`?

aperiodic:
I think Patrick's suggestion to use quantifiers rather than repeating the optional match is sound. The rest looks good.

app.rb
pcarlisle: yes

smcclellan:
Rebased and merged.

PR #300, opened 2015-04-10
Nodes collection ordering

smcclellan:


lutter:
ACK.

The Gemfile update also updates Torquebox. You should check with RE that that is ok. Now is the right time to update all the dependencies, though.

Gemfile.lock
```diff
@@ -39,56 +42,57 @@ GEM
...
-    text (1.2.3)
+      tilt (>= 1.3, < 3)
+    text (1.3.0)
     thor (0.19.1)
-    tilt (1.4.1)
-    timecop (0.6.3)
-    torquebox (3.1.1)
+    tilt (2.0.1)
+    timecop (0.7.3)
+    torquebox (3.1.2)
```
smcclellan: @melissa Is it okay to update Torquebox in the Gemfile.lock like this?

melissa: No. It's going to get mad at you for not having the right version available, since we have 3.1.1 on the builders. We need to build and ship pl-jruby to have on our builders, as well as razor-torquebox to pull into the mock during the build process

mckern:
I would absolutely not merge this without `Gemfile.lock` being removed from the PR. There's 61 version changes or Gem additions in it, with a number of brand-new additions like `addressable` or `docile` being included as top-level dependencies without corresponding updates to `Gemfile`. 

I advise moving the changes to `Gemfile.lock` to a new PR and making sure that any corresponding changes to `Gemfile` are included in the commit(s) (suggestion: don't use a pessimistic constraint on Torquebox if it's a core dependency; an explicit version is probably more desirable in this context). I'm all for moving versions forward if it's necessary but I'd prefer to keep both the PR scope constrained and the dependency history explicit and predictable.

PR #306, opened 2015-05-01
Broker error reporting

smcclellan:
This reports back information about the broker run on the node. It exposes a `log_url(msg, severity)` method which returns an endpoint that can be retrieved via a GET to log a message on the server.

This does not include Windows yet.


smcclellan:
PR updated to accommodate comments about the Powershell script.

lutter:
ACK. Looks good - I haven't tried the code, but it seems ok from reading it.

smcclellan:
Rebased, fixed spec test, and merged.

PR #308, opened 2015-06-11
(maint) Standardizing project_data.yaml, including hooks directory

smcclellan:


ext/project_data.yaml
```diff
@@ -31,19 +31,22 @@ files:
   - .bundle
   - config.ru
   - app.rb
-  - torquebox.rb
   - bin
-  - lib
+  - brokers
   - db
+  - doc
+  - ext
```
melissa: The only thing I can think of is that this might pull in the packaging artifacts in the ext directory, which we probably don't want. It looks like we already pull in the necessities from ext on lines 26-29

melissa:
:+1:

PR #309, opened 2015-06-15
Find file in directory - Needed for Hooks to work

ianlunam:
Without the slash it will look for a file called hook_type.hookcause instead of cause in the hook_type.hook directory.


smcclellan:
I'm not able to reproduce this:
```
irb(main):001:0> require 'pathname'
=> true
irb(main):002:0> Pathname.new('/abc') + 'def' + 'ghi'
=> <Pathname:/abc/def/ghi>
```
I think `+` is overridden by `Pathname`. I haven't had issues with it finding event scripts. Are you seeing different behavior?

ianlunam:
My bad. Must have been something else going on.

PR #310, opened 2015-07-08
Reinstall node with same policy

smcclellan:
Two related changes here:

- BUGFIX: Fixing deserialization of database task's boot sequence
- NEW: Allow reinstall-node with same policy

The former caused problems in the testing of the latter, so has been included in the same pull request.

### DIFF
--- a/lib/razor/data/task.rb
+++ b/lib/razor/data/task.rb
@@ -8,8 +8,20 @@ module Razor::Data
   # Note that we duck type this with Razor::Task so that they can be used
   # interchangeably for template lookup etc.
   class Task < Sequel::Model
-    plugin :serialization, :json, :boot_seq
     plugin :serialization, :json, :templates
+    # Standard json serialization doesn't work here.
+    # JSON serializes integers as strings, undo that
+    serialize_attributes [
+                         ->(b){ b.to_json },               # serialize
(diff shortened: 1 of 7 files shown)


lib/razor/data/task.rb
```diff
@@ -8,8 +8,20 @@ module Razor::Data
...
+                         ->(b){ b.to_json },               # serialize
+                         ->(b){
+                           if b.is_a?(String)
+                             b = JSON.parse(b)
+                             b.keys.select { |k| k.is_a?(String) and k =~ /^[0-9]+$/ }.
+                                 each { |k| b[k.to_i] = b.delete(k) }
+                           end
+                           b
+                         } # deserialize
+                         ], :boot_seq
```
samwoods1: Just making sure I understand...  This is an hash of values, some of which the keys are strings, and others are integers and we are just deleting any entries with an integer key from the list?

smcclellan: The `b.delete(k)` line returns the existing value, so this just reinserts the same entry with the `to_i` method called on its key.

aperiodic:
:+1:

smcclellan:
Added `NEWS.md` and `api.md` updates, rebased, and merged.

PR #312, opened 2015-07-08
Add Windows 2008 R2 Task

smcclellan:
This includes a task for installing Windows 2008 R2. It includes a few minor
changes to the generic build-razor-winpe.ps1 script.

A few workaround are/were necessary to get this to work.

- The "fastprox.dll" library, which causes issues with WinPE 4.0 + the
  2008R2 installer, is replaced at install-time by the copy included on the
  2008R2 ISO/repo.
- The machine running the build-razor-winpe.ps1 script requires the Windows
  Management Framework (WMF) version 4.0 (which requires that the Windows
  Update service be running in order to install), as well as the usual Windows
  Assessment and Deployment Kit (Windows ADK) version 8.0 or 8.1.

One other modification needed to be made to the generic Windows installer. The
samba share now uses the shell's `hostname` command, rather than the Ruby
URI library's `#hostname` command. The latter retains the domain of the URI,
which is stripped away in samba share workgroups.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-639

### DIFF
--- a/tasks/windows/2012r2.task/README.md
+++ b/tasks/windows/2012r2.task/README.md
@@ -1,5 +1,14 @@
 # Task notes for Windows 2012 R2
 
+## Pre-install Steps
+
+Follow same instructions for all other Windows tasks:
+
+- Build winpe.wim image using the `build-razor-winpe.ps1` script.
+- Copy winpe.wim to the server.
+- Make sure winpe.wim is readable by the Razor service.
+- Install/configure Samba share.
+
 ## Node Metadata
 
 - 'productkey' (optional) - This will be substituted into the unattended.xml
--- a/tasks/windows/8pro.task/README.md
+++ b/tasks/windows/8pro.task/README.md
@@ -1,5 +1,14 @@
 # Task notes for Windows 8 Professional
 
+## Pre-install Steps
+
+Follow same instructions for all other Windows tasks:
+
+- Build winpe.wim image using the `build-razor-winpe.ps1` script.
+- Copy winpe.wim to the server.
+- Make sure winpe.wim is readable by the Razor service.
+- Install/configure Samba share.
+
 ## Node Metadata
 
 - 'productkey' (optional) - This will be substituted into the unattended.xml


tasks/windows.task/second-stage.ps1.erb
```diff
@@ -16,10 +17,10 @@ $unattended = "${env:SYSTEMDRIVE}\unattended.xml"
 (new-object System.Net.WebClient).DownloadFile($unattended_src, $unattended)
 
 
-write-host "mapping SMB share \\${server}\${share} for installer access"
+write-host "mapping SMB share ${drive_path} for installer access"
 # As far as I can tell, if we don't persist, we don't connect the SMB share,
 # and that means we can't actually run stuff off it.
-new-psdrive -name 'i' -psprovider filesystem -root "\\${server}\${share}" -persist
+new-psdrive -name 'i' -psprovider filesystem -root "$($drive_path)" -persist
```
samwoods1: Was this intentional to change the path to include the ${repo} directory?  Just checking since I'm not familiar with the entire process.

tasks/windows/2008r2.task/second-stage.ps1.erb
```diff
@@ -0,0 +1,52 @@
...
+$osDrive = (bcdedit /enum OSLOADER | Select-String osdevice)[0].ToString().Split('=')[-1]
+Write-Host "Drive letter for installation is $($osDrive)"
+
+Write-Host "registering broker to run after reboot before login screen"
+New-Item -ItemType Directory -Force -Path "$($osDrive)\Windows\Setup\Scripts"
+# The $osDrive drive is the freshly installed but not-yet-booted operating system.
+(New-Object System.Net.WebClient).DownloadFile('<%= file_url('SetupComplete.cmd') %>',"$($osDrive)\Windows\Setup\Scripts\SetupComplete.cmd")
+
+write-host "restarting the system to complete installation"
+restart-computer
```
samwoods1: You could use 'restart-computer -Wait -For PowerShell -Timeout 30' to wait up to 30 seconds for the restart to happen and powershell service to be responding, instead of the start-sleep command which arbitrarily waits 30 seconds.

tasks/windows.task/second-stage.ps1.erb
smcclellan: It was intentional. I saw no reason to have the newly formed I: drive contain access to other repos. Rather than having all paths access I:/$repo/$file_path, it'll just be I:/$file_path, which I think is simpler.

tasks/windows/2008r2.task/second-stage.ps1.erb
smcclellan: Sounds fine.

build-winpe/build-razor-winpe.ps1
```diff
@@ -113,7 +112,7 @@ copy-item $wim $output
 $wim = join-path $output "winpe.wim"
 
 
-
+$env:Path = ($env:Path + ";$adk\..\..\Deployment Tools\amd64\DISM")
```
aperiodic: No need to quote or escape the space in "Deployment Tools"?

tasks/windows.task/second-stage.ps1.erb
```diff
@@ -40,5 +41,5 @@ New-Item -ItemType Directory -Force -Path "$($osDrive)\Windows\Setup\Scripts"
 (New-Object System.Net.WebClient).DownloadFile('<%= file_url('SetupComplete.cmd') %>',"$($osDrive)\Windows\Setup\Scripts\SetupComplete.cmd")
 
 write-host "restarting the system to complete installation"
-restart-computer
-start-sleep 30                  # wait for the reboot to complete before exiting
+# wait for the reboot to complete before exiting
+restart-computer -Wait -For PowerShell -Timeout 30
```
aperiodic: In general, whenever we've had a timeout like this, we've found that eventually someone will need to override it because stuff is just slow for them. Is that feasible here?

tasks/windows/2008r2.task/boot_wim.erb
```diff
@@ -0,0 +1,18 @@
...
+echo ======================================================================
+echo freeing all loaded images to ensure wimboot works
+imgfree
+echo ======================================================================
+
+set base <%= repo_url.sub(%r{/$}, '') %>
+kernel <%= file_url('wimboot', true) %>
+initrd ${base}/bootmgr				bootmgr
+initrd ${base}/boot/bcd				BCD
+initrd ${base}/boot/fonts/wgl4_boot.ttf		wgl4_boot.ttf
```
aperiodic: The alignment looks off here.

tasks/windows/2008r2.task/second-stage.ps1.erb
```diff
@@ -0,0 +1,52 @@
...
+$unattended_src = '<%= file_url('unattended.xml') %>'
+$unattended = "${env:SYSTEMDRIVE}\unattended.xml"
+(new-object System.Net.WebClient).DownloadFile($unattended_src, $unattended)
+
+write-host "fetching fastprox.dll from repo to overwrite the local copy (workaround)"
+# This is a workaround for an installer issue with WinPE 4.0 + 2008R2.
+$fastprox_loc = '<%= repo_url('sources/fastprox.dll') %>'
+$fastprox = "${env:SYSTEMDRIVE}\Windows\System32\wbem\fastprox.dll"
+$fastprox_backup = $fastprox + ".bak"
+move $fastprox $fastprox_backup
```
aperiodic: Is the original fastprox.dll guaranteed to exist? If not, what happens here?

build-winpe/build-razor-winpe.ps1
smcclellan: Since it's already in quotes and being assigned to the `env:Path` variable, it seems to be working as-is without any escaping.

aperiodic:
Done reviewing.

tasks/windows.task/second-stage.ps1.erb
smcclellan: I see no harm in bumping up the amount of time that this waits. I'll make it a minute. Any more than that, the user can always clone and modify this file as a custom task (which most places will likely want to do anyway to customize their unattended.xml).

tasks/windows/2008r2.task/boot_wim.erb
smcclellan: Will fix.

tasks/windows/2008r2.task/second-stage.ps1.erb
smcclellan: It should come bundled in the WinPE environment. There are two error cases:
- The WinPE (original) file might be absent. If this is the case, red text will appear but Powershell will continue executing. I think this is OK behavior.
- The fastprox.dll file on the server may be absent (not provided in ISO (?) or deleted by the user). If this occurs, I'm not sure we have a good fall-back for behavior. I'd expect the install to fail down the line with an error mentioning the missing file.

smcclellan:
Rebased and merged.

PR #313, opened 2015-07-09
Fix razor-server service on EL7

smcclellan:
Systemd's .service file was referencing the absent HTTP_PORT variable. This
resulted in an empty string being passed to the service for its http.port
argument. Since an empty string can't be translated to an integer, this error
resulted:

JBAS014688: Wrong type for port. Expected [EXPRESSION, INT] but was STRING

The fix is simply supplying this environment variable, which works here in
the .env file.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-629

### DIFF
--- a/ext/redhat/razor-server.env
+++ b/ext/redhat/razor-server.env
@@ -8,3 +8,4 @@ JBOSS_MODULES_SYSTEM_PKGS=org.jboss.byteman
 LAUNCH_JBOSS_IN_BACKGROUND=true
 LANG=en_US.UTF-8
 JAVA_OPTS="-Xms128m -Xmx1024m -XX:MaxPermSize=256m -Djava.net.preferIPv4Stack=true"
+HTTP_PORT=8080
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 
 ### Other
 
++ BUGFIX: The EL7 packages will now start the razor-server service properly.
 + BUGFIX: Tasks created through the `create-task` command will now find
   the correct boot stage, rather than feeding the `default` stage at each boot.
 + NEW: Task added for Windows 2008 R2. Details are on the [Wiki](https://github.com/puppetlabs/razor-server/wiki/Installing-windows).


samwoods1:
Any reason we might want to check to see if HTTP_PORT is already set before arbitrarily setting it to 8080?  I can't think why it would be anything other than that, but who knows.

smcclellan:
We may potentially want to use a more descriptive name for the variable, like `$RAZOR_HTTP_PORT` or something. For now, this matches the behavior on the PE side, so I will ticket this to fix in both places for a future PR.

ext/redhat/razor-server.env
```diff
@@ -7,4 +7,5 @@ JBOSS_CONFIG=standalone.xml
 JBOSS_MODULES_SYSTEM_PKGS=org.jboss.byteman
 LAUNCH_JBOSS_IN_BACKGROUND=true
 LANG=en_US.UTF-8
-JAVA_OPTS="-Xms128m -Xmx1024m -XX:MaxPermSize=256m -Djava.net.preferIPv4Stack=true"
\ No newline at end of file
+JAVA_OPTS="-Xms128m -Xmx1024m -XX:MaxPermSize=256m -Djava.net.preferIPv4Stack=true"
+HTTP_PORT=8080
```
aperiodic: It's a small cost to add a newline add the end of files and it's [generally recommended to avoid weird issues with some tools](http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline).

smcclellan: Fixing.

ext/redhat/razor-server.env
```diff
@@ -7,4 +7,5 @@ JBOSS_CONFIG=standalone.xml
 JBOSS_MODULES_SYSTEM_PKGS=org.jboss.byteman
 LAUNCH_JBOSS_IN_BACKGROUND=true
 LANG=en_US.UTF-8
-JAVA_OPTS="-Xms128m -Xmx1024m -XX:MaxPermSize=256m -Djava.net.preferIPv4Stack=true"
\ No newline at end of file
+JAVA_OPTS="-Xms128m -Xmx1024m -XX:MaxPermSize=256m -Djava.net.preferIPv4Stack=true"
```
aperiodic: The "preferIPv4Stack" option sounds like it'll still let you use IPv6, but in fact it [disables IPv6 entirely](http://download.java.net/jdk7/archive/b123/docs/api/java/net/doc-files/net-properties.html). Is this what you want? Will folks never want to use razor on IPv6-only hosts?

ext/redhat/razor-server.env
pcarlisle: there's an intellij setting to always do this

ext/redhat/razor-server.env
smcclellan: This is probably worth keeping in mind. I haven't heard any users bringing up any issues with this (it's been that way since the start). If this comes up as an issue, we can create a ticket and do testing without this argument.

smcclellan:
Added `NEWS.md` entry, rebased, and merged.

PR #316, opened 2015-07-29
Add support for erb templates in hostname pattern

cwood:
Instead of gsub the id this will just use the erb template. Should allow to use facts/metadata when creating host name. Also should allow customizing the id

### DIFF
--- a/lib/razor/data/node.rb
+++ b/lib/razor/data/node.rb
@@ -1,4 +1,6 @@
 # -*- encoding: utf-8 -*-
+require 'erb'
+
 module Razor::Data
   class DuplicateNodeError < RuntimeError
     attr_reader :hw_info, :nodes
@@ -207,7 +209,9 @@ module Razor::Data
       #    policy or not
       self.installed = nil
       self.root_password = policy.root_password
-      self.hostname = policy.hostname_pattern.gsub(/\$\{\s*id\s*\}/, id.to_s)
+      self.hostname = ERB.new(
+          policy.hostname_pattern.gsub(/\$\{\s*id\s*\}/, id.to_s)
+      ).result(binding)
 
       if policy.node_metadata
         modify_metadata('no_replace' => true, 'update' => policy.node_metadata)
--- a/spec/data/node_spec.rb
+++ b/spec/data/node_spec.rb
@@ -233,7 +233,7 @@ describe Razor::Data::Node do
     end
 
     it "is set from the policy's hostname_pattern when bound" do
-      policy.hostname_pattern = "host${id}.example.org"
+      policy.hostname_pattern = "host<%= id %>.example.org"
       policy.save
       node.bind(policy)
       node.hostname.should == "host#{node.id}.example.org"
--- a/spec/fabricators/fabricator.rb
+++ b/spec/fabricators/fabricator.rb
@@ -57,7 +57,7 @@ end
 Fabricator(:policy, :class_name => Razor::Data::Policy) do
   name             { Faker::Commerce.product_name + " #{Fabricate.sequence}" }
   enabled          true
-  hostname_pattern 'host${id}.example.org'
+  hostname_pattern 'host<%= @id %>.example.org'
   root_password    { Faker::Internet.password }
 
   repo
@@ -144,7 +144,6 @@ Fabricator(:bound_node, from: :node) do
   end
 
   after_save do |node, _|
-    node.hostname = node.policy.hostname_pattern.gsub('${id}', node.id.to_s)
     node.save
   end
 end


spec/data/node_spec.rb
```diff
@@ -233,9 +233,10 @@ def canonicalize(hw_info)
     end
 
     it "is set from the policy's hostname_pattern when bound" do
-      policy.hostname_pattern = "host${id}.example.org"
+      policy.hostname_pattern = "host<%= id %>.example.org"
       policy.save
       node.bind(policy)
+      puts node.id
```
smcclellan: This is debugging, I infer?

lib/razor/data/node.rb
```diff
@@ -207,7 +209,7 @@ def bind(policy)
       #    policy or not
       self.installed = nil
       self.root_password = policy.root_password
-      self.hostname = policy.hostname_pattern.gsub(/\$\{\s*id\s*\}/, id.to_s)
+      self.hostname = ERB.new(policy.hostname_pattern).result(binding)
```
smcclellan: It may be worth leaving the `gsub` in there for backward compatibility.

spec/app/create_policy_spec.rb
```diff
@@ -28,7 +28,7 @@
         'repo'          => repo.name,
         'task'          => 'some_os',
         'broker'        => broker.name,
-        'hostname'      => "host${id}.example.com",
+        'hostname'      => "host<%= id %>.example.com",
```
smcclellan: Following the backward-compatibility suggestion above, this would need to be split into two test cases.

spec/data/node_spec.rb
cwood: yea that was for debugging can be removed

lib/razor/data/node.rb
cwood: Fair enough

spec/app/create_policy_spec.rb
cwood: I dont think this is actully tested in that spec so ill just return it back.

cwood:
I updated with the gsub support and squashed the commits.

smcclellan:
Just to clarify, which attributes would be useful for the hostname that could be achieved through an ERB template?

cwood:
I would think metadata at least from a users perspective. Its also just nice to do formatting of the id. So if you wanna pad 0s you can use sprintf or however you want to format it. For my use case right now I plan on modifying the current cound of nodes bound to a policy and returning back that count so I can use the policy hostname pattern + the number so I dont have a db20 when its really the first db of that type.

ChrisPortman:
I have adjusted my tasks to read the hostname from metadata if it exists 
(and also set my policies to require it) and fall back to the hostname 
attribute.  Combining this with a hook to generate the hostname metadata 
value would probably be the best way to go about this.  You can 
implement exactly what you need without having to anticipate all the 
possible use cases that would make adding the functionality to the core 
code appropriate.  Also, when using hooks, you can have a counter of 
hook runs that can only increment, unlike policy bind count which can go 
up and down resulting in duplicated hostnames.

E.g The issue I see with depending on policy binds is:

newHost1 --> bind policyA, bind count = 1, hostname = myhost1
newHost2 --> bind policyA, bind count = 2, hostname = myhost2
newHost3 --> bind policyA, bind count = 3, hostname = myhost3
DELETE newHost2 --> policyA bind count = 2
newHost4 --> bind policyA, bind count = 3, hostname = myhost3 !Oops, 
same hostname as newHost3.

Thanks,
Chris

cwood:
yea but is it actully implemented and available? This seems like at least a good plus until thats ready.

ChrisPortman:
You mean hooks?  Yes as of 0.16.

cwood:
TBH its not even that hard to fix that issue you describe. Assuming the issue is a node2 gets deleted. You can easily scan hostnames that match and fill in with the first available...

Eg.
App01
App02
App03

App02 gets deleted

app02 gets rebound

scan hosts, increment count and verify hostname already not take, then set the count.

cwood:
Alos these are the hooks that are available according or your documentation. https://github.com/puppetlabs/razor-server/blob/master/doc/hooks.md#available-events Maybe they just need updated with the other events.

ChrisPortman:
would the node-bound-to-policy do the job?

cwood:
I think I am done. This is probably the most infuriating PR I have done.

ChrisPortman:
Sorry dude, I'm not trying to be obstructive, just looking at how you may achieve your goal now with what exists.

Dont close your PR based on my feedback.  I don't represent Puppetlabs and am not an owner of the project and have no say on whether the PR is accepted or not.

If I've missed the point, just ignore me.

cwood:
@smcclellan Any update on this? Are you good to merge this ?

smcclellan:
@cwood — I think part of the confusion in this pull request is that hooks are a feature which allow users to write their own scripts for additional functionality they may need. Nonetheless, since this is becoming quite a popular functionality request, I've submitted a pull request for a sample hook which would set a user's hostname here: https://github.com/puppetlabs/razor-server/pull/318. Can you please confirm whether this will work for your use case? Until it is merged, you could copy the included `hostname.hook` directory into your own environment and use it freely.

The primary reason for the push-back on this here is that `hostname` and `root_password` on the policy were stop-gap special fields which are now basically obsolete now that a more dynamic method of setting these fields (hooks) exists. I'd like to avoid extra logic (and extra testing, future support of complete ERB capability, etc.) around these fields if we can help it.

Thank you for the PR, though! Let me know if this solution doesn't meet your needs, or if there are any issues getting the above hook into your environment and I'll advise however I can.

cwood:
I mean I would disagree implementing the ERB template makes it so you might not even have to make your own hook. Could make it really easy for people to setup there hostnames by just using the node data via erb. Also comparitivly this change is about 8 additions vs the 300~ additions in the new hook change. I think both would be good not just forcing hooks on everyone. Seems like more overhead that may be needed in some cases.

PR #317, opened 2015-07-29
Get current policy count

cwood:
This is a added feature on top of #316 that will get the current number of nodes attached to this policy so you can use this in your hostname erb template. This is nice in that each policy declares its hostname and I want to use the count to setup app1, and db1, but have them as node1, and node2.

### DIFF
--- a/lib/razor/data/node.rb
+++ b/lib/razor/data/node.rb
@@ -1,4 +1,6 @@
 # -*- encoding: utf-8 -*-
+require 'erb'
+
 module Razor::Data
   class DuplicateNodeError < RuntimeError
     attr_reader :hw_info, :nodes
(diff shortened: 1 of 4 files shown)


ChrisPortman:
How does this handle scenarios where nodes are unbound/deleted dropping the bind count?  Would that open up the potential for duplicate hostnames?

cwood:
Funny you say that I just ran into that issue. I need to deal with race conditions in this. But should probably also add the function to fill in counts.

smcclellan:
I'd intended to solve problems like this using hooks. I [created a ticket](https://tickets.puppetlabs.com/browse/RAZOR-642) to create a sample hook which would demonstrate dynamic hostname setting. The same race condition could still occur if you reference the policy, however. To mitigate this, it'd be best to keep those counts in the hook configuration (still not completely race-proof, but I would expect less time using stale data).

cwood:
That would be fine too just didnt see any work on it yet. I am using this in my companies instance of razor. Ill probably leave this alone then. The race conditions are still kind of bad. Ill probably leave this alone until that hook is in razor.

PR #318, opened 2015-08-05
Hostname hook

smcclellan:
One frequently requested feature with hooks is the ability to dynamically
update the hostname for a node. The sample hook being added here
will demonstrate how to do that by updating the node's `hostname` metadata
value. Then, the task can refer to this value in its templates in order to
set the hostname.

The hook operates by keeping track of the number of nodes bound to each
policy. The counter for the policy, stored as hook configuration, starts at 1.
When a node binds to this policy, the hook increments its `counter`
configuration and returns a rendered hostname.

More details are included in the file's README.md in the hostname.hook
directory.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-642
Fixes https://tickets.puppetlabs.com/browse/RAZOR-643


lutter:
ACK. Looks good to me; nice find with using Bundler.with_clean_env.

One minor nit: It would be good if the commit "Use separate queue for hooks" explained
  what the problem was with having hooks in the same queue as other jobs.

smcclellan:
Fixed per comments, rebased, merged.

PR #320, opened 2015-10-07
(maint) We should not be checking in Gemfile.lock

melissa:
Generally this makes things challenging. We have to explicitely remove
Gemfile.lock prior to building. This is an extra step we shouldn't have
to do. Maintaining Gemfile.lock also goes against standard practices in
our other project repos.


melissa:
i.e., we won't have to have https://github.com/puppetlabs/razor-server/blob/master/Rakefile#L151 anymore

PR #322, opened 2015-10-14
Add support for bypassing authentication, if request is from localhost.

mamba:
Puppet modules currently available for configuring razor itself do not support authentication.
This changes add support to enable a bypass option for requests to /api from localhost.

Example puppet module that can be used with this option.

Access from remote to the /api still needs authentication.
Example use case:
External application that communicates with razor.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -82,13 +82,18 @@ and requires full control over the database (eg: add and remove tables):
       org.apache.shiro.SecurityUtils.subject
     end
 
+    # Check if request is from localhost and if bypass for localhost is enabled
+    def local_request?
+      request.ip == '127.0.0.1' and Razor.config['auth.allow_localhost']
+    end
+
     # Assert that the current user has (all of) the specified permissions, and
     # raise an exception if they do not.  We handle that exception generically
     # at the top level.
     #
     # If security is disabled then this simply succeeds.
     def check_permissions!(*which)
-      Razor.config['auth.enabled'] and user.check_permissions(*which)
+      Razor.config['auth.enabled'] and not local_request? and user.check_permissions(*which)
       true
     end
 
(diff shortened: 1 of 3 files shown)


app.rb
```diff
@@ -82,13 +82,18 @@ def user
       org.apache.shiro.SecurityUtils.subject
     end
 
+    # Check if request is from localhost and if bypass for localhost is enabled
+    def local_request?
+      request.ip == '127.0.0.1' and Razor.config['auth.allow_localhost']
```
smcclellan: Is it possible to spoof this value in the request headers?

mamba: The IP is coming from the ENV REMOTE_ADDR and is set by the webserver.
Therefore spoofing from an external system is not possible.

Depending on the setup HTTP_X_FORWARDED_FOR is considered as well.
(https://github.com/rack/rack/blob/master/lib/rack/request.rb#L273).

Rails is just using the REMOTE_ADDR header.
(http://api.rubyonrails.org/classes/ActionDispatch/Request.html#method-i-local-3F).

sandroBeffa:
Will this be merged soon ? This is a feature we would need!

app.rb
```diff
@@ -82,13 +82,18 @@ def user
...
+    end
+
     # Assert that the current user has (all of) the specified permissions, and
     # raise an exception if they do not.  We handle that exception generically
     # at the top level.
     #
     # If security is disabled then this simply succeeds.
```
smcclellan: Can we update this comment to reflect the new behavior?

smcclellan:
Can we add some spec tests for this functionality? After that, I see no issues merging this once it is rebased.

smcclellan:
Merged as https://github.com/puppetlabs/razor-server/pull/344.

PR #324, opened 2015-11-02
Update README.md

MFredette:
added some edits.

### DIFF
--- a/hooks/hostname.hook/README.md
+++ b/hooks/hostname.hook/README.md
@@ -1,49 +1,45 @@
 ## Sample counting hook
 
+You can use a hook to create more advanced dynamic hostnames than the simple incremented pattern --- `$\{id\}.example.com` --- from the `hostname` property on a policy. This type of hook calculates the correct hostname and returns that hostname as metadata on the task. To do so, it uses a basic counter system that stores how many nodes have bound to a given policy.
+
 ### Prerequisites
 
-This hook requires that ruby is installed. This must be contained in $PATH in
-order for the hook script to succeed.
+Ruby must be installed in `$PATH` for the hook script to succeed.
 
-### Purpose
+### Install the Hook
 
-This hook performs dynamic hostname assignment. It uses a basic counter system,
-storing how many nodes have bound to a given policy.
+This hook comes with Razor. To use it, create an instance of
+the hook with the following command:
 
-### How to install
 
-This hook comes with Razor; all you will need to do is create an instance of
-the hook via:
+        razor create-hook --name some_policy_hook --hook-type hostname \
+            --configuration policy=some_policy \
+            --configuration hostname-pattern='${policy}${count}.example.com'
 
-```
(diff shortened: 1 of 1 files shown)


hooks/hostname.hook/README.md
```diff
@@ -1,49 +1,45 @@
 ## Sample counting hook
```
smcclellan: While we're editing this file, can we change this to say:
## Sample hostname assignment hook

hooks/hostname.hook/README.md
```diff
@@ -1,49 +1,45 @@
...
 
+You can use a hook to create more advanced dynamic hostnames than the simple incremented pattern --- `$\{id\}.example.com` --- from the `hostname` property on a policy. This type of hook calculates the correct hostname and returns that hostname as metadata on the task. To do so, it uses a basic counter system that stores how many nodes have bound to a given policy.
```
smcclellan: After this paragraph, let's add:
"This hook is intended to be extended for cases where an external system needs to be contacted to determine the correct hostname. In such a scenario, the new value will still be returned as metadata for the node."

Also, I think it should say:
"...returns that hostname as metadata on the node."

hooks/hostname.hook/README.md
```diff
@@ -1,49 +1,45 @@
...
+
 ### Prerequisites
 
-This hook requires that ruby is installed. This must be contained in $PATH in
-order for the hook script to succeed.
+Ruby must be installed in `$PATH` for the hook script to succeed.
```
smcclellan: Add:
"If it is not included, add it via the `hook_execution_path` property in the config.yaml file."

hooks/hostname.hook/README.md
```diff
@@ -1,49 +1,45 @@
...
-   - It replaces `${count}` with the current value of the `counter` hook 
+   - It replaces `${count}` with the current value of the `counter` hook
      configuration.
-   - It left-pads the `${count}` with `padding` zeroes. E.g. If the hook
+   - It left-pads the `${count}` with `padding` zeroes. For example, if the hook
      configuration's `padding` equals 3, a `count` of 7 will be rendered as
      `007`.
    - It replaces `${policy}` with the name of its policy.
-6. The hook then returns the rendered hostname-pattern as the node metadata
+6. The hook then returns the rendered hostname pattern as the node metadata
```
smcclellan: `hostname-pattern` is the name of the field, might be best to just add tick marks (`).

hooks/hostname.hook/README.md
```diff
@@ -1,49 +1,45 @@
...
    of `hostname`.
 7. The hook also returns the incremented value for the counter that was used.
```
smcclellan: "The hook also returns the incremented value for the counter that was used so that the next execution of the hook will use the next value."

smcclellan:
Looks great after those edits; I will merge when resolved.

smcclellan:
Rebased and merged.

PR #325, opened 2015-11-03
Update store_ip shell script to check broadcast interface

mattkirby:
This commit updates the store_ip shell script used to get the primary IP address to dynamically determine the primary interface based on what is marked for broadcast by ip. Without this change the script fails to get the IP address for a CentOS 7 host due to a hard coded eth0 value. This has been tested on CentOS and Debian and provides the primary broadcast IP address in both cases.


smcclellan:
Looks good.

PR #327, opened 2015-11-13
(maint) remove apt_repo_path and yum_repo_path from build_defaults

underscorgan:
Razor-server doesn't ship anywhere special, so we should let it pick up
the TEAM defaults.


smcclellan:
:+1:

PR #330, opened 2015-12-11
Include shebang line in /etc/rc.d/rc.local

smcclellan:
On installs with systemd, the kickstart file will drop the /etc/rc.d/rc.local
file without a shebang line at the start. This causes, in some cases, the
post_install script not to run.

This fixes the simple `touch` command to instead drop a shebang line.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-754


smcclellan:
The real root of this ticket is captured and fixed in https://github.com/puppetlabs/razor-server/pull/331. This PR may still be helpful for standards compliance.

PR #331, opened 2015-12-12
Add Fedora 23 task

smcclellan:
This adds a task for Fedora 23. Fedora releases happen quickly, so this might
only be relevant for a year, but these same steps should be valid for future
releases as well.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-755

### DIFF
--- a/tasks/fedora.task/README.md
+++ b/tasks/fedora.task/README.md
@@ -1,0 +1,13 @@
+# Task notes for Fedora 23
+
+## Node Metadata
+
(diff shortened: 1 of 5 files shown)


smcclellan:
This installs successfully with [the Fedora 23 Server ISO](http://download.fedoraproject.org/pub/fedora/linux/releases/23/Server/x86_64/iso/Fedora-Server-DVD-x86_64-23.iso).

smcclellan:
Rebased and merged: 0a0e042

PR #332, opened 2015-12-12
Fix event ordering

scottsolowild:
Facts, whether events, node log, hook log, or commands, were counter-
intuitively ordered. This commit puts the latest entry at the bottom,
which should be most useful.

Additionally, whenever a limited "fact" collection is requested, it will
limit from the bottom up. This means `razor events --limit 5` will show the 5
most recent events, with the most recent at the bottom.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-734

### DIFF
--- a/lib/razor/data/hook.rb
+++ b/lib/razor/data/hook.rb
@@ -149,8 +149,15 @@ class Razor::Data::Hook < Sequel::Model
   end
 
   def log(params = {})
-    cursor = Razor::Data::Event.order(:timestamp).order(:id).reverse.
-        where(hook_id: id).limit(params[:limit], params[:start])
+    cursor = Razor::Data::Event.order(:timestamp).order(:id).
+        where(hook_id: id)
+    total = cursor.count if cursor.respond_to?(:count)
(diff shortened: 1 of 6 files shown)


lib/razor/data/hook.rb
```diff
@@ -149,8 +149,15 @@ def validate
...
   def log(params = {})
-    cursor = Razor::Data::Event.order(:timestamp).order(:id).reverse.
-        where(hook_id: id).limit(params[:limit], params[:start])
+    cursor = Razor::Data::Event.order(:timestamp).order(:id).
+        where(hook_id: id)
+    total = cursor.count if cursor.respond_to?(:count)
+    if params[:start].nil? and params[:limit] and !total.nil?
+      # We have a request for a limited list of facts without a starting
+      # value. Take from the end so the latest entries are included.
+      params[:start] = [total - params[:limit], 0].max
```
smcclellan: This may be considered a break of the API. Probably a better way to deal with this is to add metadata properties to the previous endpoint or something. I'm not clear on the best approach here.

PR #333, opened 2015-12-15
Delete repo directory when repo is deleted

smcclellan:
When the user runs `razor delete-repo --name some-repo`, the intent is to
delete the folder if the user supplied an `iso-url` property. That was, in many
cases, not happening because many ISO archives contain files which are
read-only, so the secure delete attempt fails.

The fix for this is to update the contents of the repo directory to allow write
privileges so they can be deleted properly.

This manifested when the user tried to create a new repo with the same name.
The archive extraction would fail from libarchive, and the error was "Can't
unlink already-existing object", likely from the same permissions reason.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-757

### DIFF
--- a/lib/razor/data/repo.rb
+++ b/lib/razor/data/repo.rb
@@ -34,6 +34,13 @@ module Razor::Data
       end
     end
 
+    def remove_directory(dir)
+      if Dir.exist?(iso_location)
+        FileUtils.chmod_R('+w', dir, force: true)
+        FileUtils.remove_entry_secure(dir)
+      end
+    end
+
     # When we are destroyed, if we have a scratch directory, we need to
     # remove it.
     def after_destroy
@@ -46,7 +53,9 @@ module Razor::Data
 
       # Remove repo directory.
       if Dir.exist?(iso_location)
-        FileUtils.remove_entry_secure(iso_location, true)
+        # Some files in archives are write-only. Change this property so the
+        # delete succeeds.
+        remove_directory(iso_location)
       end
     end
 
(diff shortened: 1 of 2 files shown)


steveax:
I get 4 test failures:

````
Failures:

  1) command and query API /api/collections/repos should list all repos
     Failure/Error: repos.size.should == 2
       expected: 2
            got: 3 (using ==)
     # ./spec/app/api_spec.rb:333:in `(root)'

  2) command and query API /api/collections/commands with none it should behave like a command collection should return a valid collection
     Failure/Error: nodes.count.should == expected
       expected: 0
            got: 1 (using ==)
     Shared Example Group: "a command collection" called from ./spec/app/api_spec.rb:964
     # ./spec/app/api_spec.rb:943:in `(root)'

  3) command and query API /api/collections/commands with one it should behave like a command collection should return a valid collection
     Failure/Error: nodes.count.should == expected
       expected: 1
            got: 2 (using ==)
     Shared Example Group: "a command collection" called from ./spec/app/api_spec.rb:972
     # ./spec/app/api_spec.rb:943:in `(root)'

  4) command and query API /api/collections/commands with ten it should behave like a command collection should return a valid collection
     Failure/Error: nodes.count.should == expected
       expected: 10
            got: 11 (using ==)
     Shared Example Group: "a command collection" called from ./spec/app/api_spec.rb:980
     # ./spec/app/api_spec.rb:943:in `(root)'

Finished in 58.41 seconds
2315 examples, 4 failures
````

smcclellan:
@steveax It looks like that would have occurred if you ran `create-repo` on the database prior to running the tests. I would try resetting the database (`razor-admin reset-database`) and rerunning the tests.

steveax:
Yep, that was it. Sorry, used to the test DB being isolated. LGTM!

PR #334, opened 2015-12-18
Escape unsafe characters in kickstart

smcclellan:
In cases where interpolation occurs in a kickstart file, special characters
could be rendered which change the meaning of the file or make it invalid. This
wraps such problematic cases in a sanitizing block to prevent such issues from
occurring.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-205


smcclellan:
This method doesn't quite work for some reason. Further testing and iteration is required.

smcclellan:
Closing until a working approach forward is determined.

PR #335, opened 2015-12-18
Fix delete-repo without iso-url property

smcclellan:
If a repo is created using `no-content`, we will assume that the user has
added files to the repo directory and that this directory should remain as-is,
not being deleted when the repo is deleted.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-757


steveax:
LGTM!

PR #337, opened 2016-01-14
Removed apt-get commands from os_boot.erb

jimmypw:
The os_boot script is run at every boot. This should never be upgrading the os it undermines the administrator.

### DIFF
--- a/tasks/debian.task/os_boot.erb
+++ b/tasks/debian.task/os_boot.erb
@@ -5,11 +5,6 @@ exec >> /var/log/razor.log 2>&1
 <%= render_template("set_hostname") %>
 
 sed -i 's_<%= repo_url("archive.ubuntu.com/ubuntu_g")%>' /etc/apt/sources.list
-apt-get -y update
-[ "$?" -eq 0 ] || curl -s <%= log_url("apt_update_fail", :error) %>
-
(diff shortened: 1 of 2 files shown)


smcclellan:
`os_boot` should be rendered into the preseed file under the `preseed/late_command` callback. Is that not the behavior you're seeing?

If it only runs once after an install completes, is that an appropriate time for upgrading?

jimmypw:
That is correct, the late preseed file is inserting the os_boot script and inserting a call to run in rc.local - that will call it on every boot.

There is however no locking around the file preventing it to be run for a second time. I can appriciate that if this file is run on first boot only you would want to do a full upgrade.

darthmdh:
It only runs once, os_complete.erb removes the call of the postinst script from rc.local.

smcclellan:
@jimmypw Is this the behavior you were seeing? I believe darthmdh is correct in that the os_complete.erb script should prevent the apt-get commands from running again.

smcclellan:
I will close this PR for now due to inactivity. If the issue (or comment above) isn't the observed behavior, feel free to open a new PR and/or [file a ticket](https://tickets.puppetlabs.com/browse/RAZOR).

PR #338, opened 2016-02-08
(RAZOR-779) match (but do not bind) when protect_new_nodes is true

steveax:
Prior to this commit, when protect_new_nodes was set to true, nodes
would not be matched or bound. This commit changes the behavior so
that new nodes will always be matched, even if they are not bound
due to protect_new_nodes being set to true.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-779

### DIFF
--- a/lib/razor/data/node.rb
+++ b/lib/razor/data/node.rb
@@ -273,14 +273,18 @@ module Razor::Data
       (new_tags - self.tags).each { |t| self.add_tag(t) }
     end
 
-    # Update the tags for this node and try to bind a policy.
-    def match_and_bind
+    # Update the tags for this node.
+    def match_tags
       eval_tags
-      Policy.bind(self)
     rescue Razor::Matcher::RuleEvaluationError => e
       log_append :severity => "error", :msg => e.message
       save
       raise e
+    end
+
+    # Try to bind a policy to this node.
+    def bind_policy
+      Policy.bind(self)
     end
 
     # Modify metadata the API reciever does alot of sanity checking.
@@ -338,7 +342,8 @@ module Razor::Data
       # time, i.e. have the update statement do 'last_checkin = now()' but
       # that is currently not possible with Sequel
       self.last_checkin = Time.now
-      match_and_bind unless (installed or policy)
+      match_tags
+      bind_policy unless (installed or policy)
       if policy
         log_append(:action => :reboot, :policy => policy.name)
       elsif installed
--- a/spec/app/provisioning_spec.rb
+++ b/spec/app/provisioning_spec.rb
@@ -288,7 +288,7 @@ describe "provisioning API" do
       last_response.status.should == 404
     end
   end
-  
+
   describe "storing node metadata" do
     before(:each) do
       @node = Fabricate(:node)
@@ -512,8 +512,6 @@ describe "provisioning API" do
         node.registered?.should be_true
         node.installed.should == installed
         node.policy.should be_nil
-        # Checkin will not try to evaluate tags on installed nodes
-        node.tags.should be_empty
       end
     end
   end
--- a/spec/data/node_spec.rb
+++ b/spec/data/node_spec.rb
@@ -275,6 +275,35 @@ describe Razor::Data::Node do
     end
   end
 
+  describe "matching on checkin" do
+    hw_id = "001122334455"
+
+    let (:tag) {
+      Tag.create(:name => "t1", :matcher => Razor::Matcher.new(["=", ["fact", "f1"], "a"]))
+    }
+    let (:node) {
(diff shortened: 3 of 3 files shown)


lib/razor/data/node.rb
```diff
@@ -273,9 +273,17 @@ def eval_tags
...
+    rescue Razor::Matcher::RuleEvaluationError => e
+      log_append :severity => "error", :msg => e.message
+      save
+      raise e
+    end
+
+    # Try to bind a policy to this node.
+    def bind_policy
       Policy.bind(self)
     rescue Razor::Matcher::RuleEvaluationError => e
```
smcclellan: This is only relevant for `match_tags` now.

spec/app/provisioning_spec.rb
```diff
@@ -512,8 +512,6 @@ def checkin(node, facts)
         node.registered?.should be_true
         node.installed.should == installed
         node.policy.should be_nil
-        # Checkin will not try to evaluate tags on installed nodes
-        node.tags.should be_empty
```
smcclellan: Can this instead test that the tag is on the node?

spec/data/node_spec.rb
```diff
@@ -331,6 +331,21 @@ def canonicalize(hw_info)
       node.installed.should == 'test'
     end
 
+    it "should update tags if the node is marked installed" do
+      node.installed = 'test'
+      node.save
+
+      policy = Fabricate(:policy, :rule_number => 20)
```
smcclellan: Testing this shouldn't require a policy at all. Tags are matched regardless of which policies exist.

spec/data/node_spec.rb
```diff
@@ -331,6 +331,21 @@ def canonicalize(hw_info)
...
+      policy.add_tag(tag)
+      policy.save
+
+      node.checkin('facts' => { 'f1' => 'b' })
+
+      node.reload
+      node.policy.should be_nil
+      node.installed.should == 'test'
```
smcclellan: I don't see a check that it "should update tags" in this test.

spec/data/node_spec.rb
```diff
@@ -331,6 +331,21 @@ def canonicalize(hw_info)
       node.installed.should == 'test'
     end
 
+    it "should update tags if the node is marked installed" do
```
smcclellan: I'd actually move this test out of "binding on checkin" since this does not test binding at all.

spec/data/node_spec.rb
```diff
@@ -366,13 +382,13 @@ def make_tagged_policy(sort_order)
         policy20 = make_tagged_policy(20)
         node.checkin("facts" => { "f1" => "a" })
         node.reload
-        node.tags.should be_empty
         node.policy.should == random_policy
       end
 
       it "should not change when a tag changes" do
```
smcclellan: This title would need to be updated, as written: `"should update tags"` or similar.

spec/data/node_spec.rb
```diff
@@ -382,8 +398,8 @@ def make_tagged_policy(sort_order)
 
         node.checkin("facts" => { "f1" => "a" })
         node.reload
-        # node.tags reflects the tags that applied when the node was bound
-        node.tags.should == [ tag ]
+        # node.tags should be updated to be and empty set since the match is false
```
smcclellan: Typo here.

spec/data/node_spec.rb
smcclellan: Better would be under the `protect_new_nodes` section, perhaps.

spec/data/node_spec.rb
smcclellan: I would expect a tag to be created here, rather than a policy.

smcclellan:
Done reviewing. Can you add `Fixes https://tickets.puppetlabs.com/browse/RAZOR-779` to the end of this commit? It makes it easier for referencing than just the ticket number in the title.

spec/data/node_spec.rb
```diff
@@ -275,6 +275,29 @@ def canonicalize(hw_info)
...
+      Node.create(:hw_info => ["mac=#{hw_id}"], :facts => { "f1" => "a" })
+    }
+
+    it "should update tags if the node is marked installed" do
+      node.installed = 'test'
+      node.save
+      node.match_tags
+      tag.match?(node).should be_true
+
+      node.checkin('facts' => { 'f1' => 'a' })
```
smcclellan: The main goal of this test is to verify that this call here is changing the tags on a node marked as installed. Can you add a check prior to this which verifies the previous state of the node's tags as different from the later state of the node's tags?

steveax: Yep, will do.

smcclellan:
:+1:

PR #341, opened 2016-03-07
Add Puppet-AIO Broker.

jmkeyes:
This PR adds an additional broker that deploys the Puppet AIO package instead of the standard Puppet agent. It's almost virtually identical to the standard Puppet broker; a cleaner approach may be to add an additional flag or variable to the standard Puppet broker to use the AIO package instead.

### DIFF
--- a/brokers/puppet-aio.broker/configuration.yaml
+++ b/brokers/puppet-aio.broker/configuration.yaml
@@ -1,0 +1,7 @@
+---
+certname:
+  description: "The locally unique name for this node."
+server:
+  description: "The puppet master server to request configurations from."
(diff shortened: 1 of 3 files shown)


smcclellan:
@jmkeyes Thanks for the pull request! I've taken your work and adapted it to be more of a migration, so the old `puppet` broker will be `legacy-puppet` and the new `puppet-aio` broker in this PR will be simply `puppet`.

I'm closing this PR in favor of another PR (https://github.com/puppetlabs/razor-server/pull/364), but have retained credit for the code you wrote in the commit. Cheers!

PR #347, opened 2016-05-27
(maint) Change razor pathing

melissa:


melissa:
Required for https://github.com/puppetlabs/razor-vanagon/pull/1

melissa:
Corresponds to https://github.com/puppetlabs/puppet-specifications/pull/77

bin/razor-admin
```diff
@@ -78,7 +86,7 @@ end
 
 # It is very important that razor/initialize isn't loaded until after
 # RACK_ENV has been set
-require_relative '../lib/razor/initialize'
+require 'razor/initialize'
```
smcclellan: Why lose the relative require?

melissa: When referencing a file from a symlink, the relative path was being evaluated from the symlink location rather than the source location, which meant the libraries couldn't be found. Rather than loading a relative library, it was easier to add the razor libs directory to the Ruby load path. That, and @mckern said something about `required_relative` being deprecated

lib/razor/config.rb
```diff
@@ -43,7 +43,7 @@ def initialize(env, fname = nil, defaults_file = nil)
       # Use the filename given, or from the environment, or from /etc if it
       # exists, otherwise the one in our root directory...
       fname ||= ENV["RAZOR_CONFIG"] ||
-        (File.file?('/etc/razor/config.yaml') and '/etc/razor/config.yaml') ||
+        (File.file?('/etc/puppetlabs/razor-server/config.yaml') and '/etc/puppetlabs/razor-server/config.yaml') ||
```
smcclellan: Will packaging move the old config.yaml files into place during upgrade somehow? Otherwise it looks like the user will need to move their config to this new location manually.

bin/razor-admin
smcclellan: Ok, just curious.

melissa: 👍  let me know if it's better to back out this change

lib/razor/config.rb
melissa: As it stands, no.

There are pathing changes required to the config file for it to work (see above). So, what the packaging does it take the updated config.yaml and drops it into place. I was hoping we would update the docs to suggest a partial migration where the user effectively ensures any updates to their config file are moved over to the new one. We could probably move the old file and just sed out any of the pathing updates we know need to happen. I'm not sure when I'll have time to take care of that though

bin/razor-admin
mckern: I can't find what I thought I read about it being deprecated, but there is some pretty solid guidance about when to use it and when not to use [it in this thread](https://bugs.ruby-lang.org/issues/8781).  Maybe I misremembered and conflated it with `autoload`. If I did, that's on me. Sorry for the confusion.

mckern: Oh no, I remember now. [Jruby implemented require_relative slightly differently than MRI](https://github.com/jruby/jruby/issues/3092). It's probably fixed now, but it was squishy behavior in the spec that led to my thoughts on not using require relative.

lib/razor/config.rb
smcclellan: I think either approach is good, but I'd prefer the invisible-to-user approach if it's not too difficult. Maybe if the config.yaml file is in a different location, or for some reason we cannot move it, then we just instruct the user on how to copy those changes over.

melissa: That's a change that we'll have to make in the razor-vanagon repo, not here. That doesn't help with people who have installed from source though.

melissa: I don't think this PR should be blocked on that decision, since it is a packaging change. What do you think, @smcclellan ?

smcclellan: Sounds good. I will merge this PR.

smcclellan:
This PR required spec test updates found in this commit: https://github.com/puppetlabs/razor-server/commit/8a0a155a4a02f3409deafbe5f139fd9f9e271acb

melissa:
Thank you for catching and taking care of that!

PR #348, opened 2016-06-03
Changes to the second-stage powershell script

jimmypw:
- Parameterised the drive letter that is used to host the windows install files
- Changed the location that razor will look for the install files in samba.
  - I have changed this because of very inconsistant results when attempting
    to map the drive letter to a folder within a share. It would fail roughly
    7 in 10 times. The change means that you must create a samba share for each
    repository that you will expose over samba the trade off is a more reliable
    install.

This change if accepted will require a change to the documentation on https://github.com/puppetlabs/razor-server/wiki/Installing-windows#create-smb-share. The samba configuration now looks like the following.

[windowssvr2012r2]
  comment   = Windows Server 2012 R2 Install
  # this is, by default, under /var/lib/razor/repo-store/
  path      = /var/lib/razor/repo-store/windowssvr2012r2
  guest ok  = yes
  writable  = no
  browsable = yes

I also found some additional settings that seem to make the connection to the samba share somewhat more reliable these settings are:

[global]
  kernel oplocks = no
  nt acl support = no
  strict locking = no

### DIFF
--- a/tasks/windows.task/second-stage.ps1.erb
+++ b/tasks/windows.task/second-stage.ps1.erb
@@ -6,10 +6,10 @@ write-host "Razor second-stage Windows installer script starting"
 # installer.  That will, I guess, have to change some time soon.  Look to see
 # something akin to the way that brokers work implemented.
 $server = '<%= `hostname`.strip %>'
-$share  = 'razor'
 $repo   = '<%= repo_url.split('/').last %>'
-$drive_path = "\\${server}\${share}\${repo}"
-$installer = "${drive_path}\setup.exe"
+$drive_path = "\\${server}\${repo}"
+$drive_letter = "I"
+$installer = "${drive_letter}:\setup.exe"
 
 write-host "fetching the unattended.xml file to feed the installer"
 $unattended_src = '<%= file_url('unattended.xml') %>'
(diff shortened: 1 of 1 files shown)


smcclellan:
Hi @jimmypw, thanks for the PR. Can you create a [JIRA ticket](https://tickets.puppetlabs.com/browse/RAZOR) for the WinPE issue that this PR addresses, so we can track testing of this fix in the product?

smcclellan:
@jimmypw Are you still seeing inconsistencies with this? Here are thoughts on the three changes discussed in this PR:

- I'm happy to add the three samba options to our recommended config.
- Parameterizing the drive letter makes sense. I don't see how the drive letter would be set from this, though?
- I have reservations around forcing the user to create a new samba share for every new repository. That sounds like a lot of overhead to me. I'd prefer to keep the path the same.

Nekototori:
I was also seeing inconsistencies, specifically on 2016.  Digging around, it looks like some needed services weren't enabling, and the PS-Drive command sometimes required username/pass to connect successfully when hosting tftp on razor (and using the pe_razor class to configure it).  I added the following lines to my build-razor-winpe.ps1 script, but it could easily be adapted to the second-stage.ps1.  Even though the samba share is set to everyone has access, short of modifying the registry of my new build to reduce security, I need to pass *something* in to the user/pass field, so I just pass x for username and y for password.  No legitimate credentials required.

`write-host "* Writing Windows\System32\startnet.cmd script"
    $file = join-path $mount "Windows\System32\startnet.cmd"
    set-content $file @"
    @echo off
    echo starting wpeinit to detect and boot network hardware
    wpeinit
    echo telling wpe to initialize network
    wpeutil initializenetwork
    echo doing some fixes
    ping -n 15 127.0.0.1
    net use i: \\razor\razor\win2016 /user:x y
    echo starting the razor client
    powershell -executionpolicy bypass -noninteractive -file %SYSTEMDRIVE%\razor-client.ps1
    echo dropping to a command shell now...
"@`

jimmypw:
Hey, sorry i didn't realise this was still open. I don't know if it is still an issue I ended up ripping it out and doing my own thing.

PR #350, opened 2016-06-29
(maint) Move all service files to toplevel ext dir

melissa:
This commit also removes all obsolete packaging artifacts. Razor server
packages should be built with vanagon using the razor-vanagon project.


smcclellan:
👍

PR #354, opened 2016-08-23
Add `update-policy-repo` to update a policy's repo

smcclellan:
Two feasible use cases for needing a command like this are:
- A user notices a mistype of the source for a repo after creating it.
- A user decides that a policy should use a different repo (OS update without
  messing with the policy list)

Updating a policy's repo will provide this functionality with little
interruption to existing installs, if any are occurring. This command should
simply update the reference so that repo file requests can be immediately
redirected to the new repo.

The command will use positional arguments of `policy` then `repo`.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-762

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -386,6 +386,19 @@ policy when you run this command, provisioning errors may occur.
     "task": "other_task"
     }
 
+### Update policy repo
+
+This ensures that a specified policy uses the repo this command specifies.
+If a node is currently provisioning against the
+policy when you run this command, provisioning errors may occur.
+
+    The following shows how to update a policy’s repo to a repo called “other_repo”.
+
+    {
+    "policy": "my_policy",
+    "repo": "other_repo"
+    }
+
 ### Delete policy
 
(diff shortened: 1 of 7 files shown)


doc/api.md
```diff
@@ -388,6 +388,20 @@ policy when you run this command, provisioning errors may occur.
...
+### Update policy repo
+
+This ensures that a specified policy uses the repo this command specifies.
+If a node is currently provisioning against the
+policy when you run this command, provisioning errors may occur.
+
+    The following shows how to update a policy’s repo to a repo called “other_repo”.
+
+    {
+    "node": "node1",
```
steveax: node seems to be an errant property?

```
curl -X POST -H 'Content-Type: application/json' \
> -d '{
  "node": "node1",
  "policy": "my_policy",
  "repo": "fedora20"
  }' \
   http://localhost:8080/api/commands/update-policy-repo
{"error":"extra attribute node was present in the command, but is not allowed"}%
```

steveax:
image
LGTM!

PR #355, opened 2016-08-23
Synchronize time in puppet and puppet-pe brokers

smcclellan:
Without running `ntpdate` before registering with the puppet master, sometimes,
the agent can't retrieve a catalog with an error that says the certificate is
not yet valid.

In most cases, a network will already be configured to use `ntpdate`. A node
may, however, need to set this value before checking in with the puppet master.

To make this as generic as possible, both brokers now include a
new, optional `ntpdate_server` config. Without this config present, no
`ntpdate` command will occur.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-933

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,12 @@
 + IMPROVEMENT: Added `update-policy-broker` command to facilitate
   migrating the broker that a policy uses.
 
+### Other
+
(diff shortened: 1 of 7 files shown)


steveax:
👍

PR #356, opened 2016-08-31
Add `update-policy-broker` command

smcclellan:
The current method for migrating a policy to use a different broker involves
deleting and recreating the policy. Since we will want to update stock brokers
in the future, we need to supply users a way to change their broker without
deleting policies first. This command will do that.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-935


steveax:
👍

PR #360, opened 2016-09-29
Add `win_language` metadata used by Windows installers

smcclellan:
In order for our stock Windows task to allow ISO files in languages other
than `en-US`, users previously needed to override the entire unattended.xml.erb
file.

In the interest of making the stock tasks more generic, this should
implement a "metadata-with-default" pattern. If the node has the `win_language`
metadata, that will be used as the UILanguage (which causes the error), the
InputLocale, the SystemLocale, and the UserLocale.

When a user wishes to install an ISO in a language other than `en-US`, the
recommended approach is to add the `win_language` metadata to the node via the
`node_metadata` property on the given Windows policy.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-938


steveax:
👍

PR #361, opened 2016-10-07
Node Metadata updates

smcclellan:
This PR fixes two related issues: RAZOR-672 and RAZOR-940.

#1:
Users may need to modify the node metadata that gets assigned when a node binds
to a policy. There is currently no command to perform this, so the user must
work around the issue by either deleting the policy or creating a specialized
hook to achieve this.

This commit creates a new command, `update-policy-node-metadata`, to handle
this case.

#2:
When the user runs either `modify-node-metadata` or `update-node-metadata`
using the `no_replace` flag, the result is that the operation fails silently
for any key(s) that already exist. This is not the behavior documented for
these commands, which state that the command will fail if this happens.

In the case of `modify-node-metadata`, a new flag `force` will be added so that
batch operations can still succeed. This will allow feature parity with the
existing behavior by supplying a value of `true`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,9 @@
   exists. For the latter command, you may supply the new argument
   `force` to achieve the old functionality of simply skipping the
   replacing key.
++ NEW: Added `update-policy-node-metadata` command to facilitate
+  changing the metadata that gets added to a node when it binds to a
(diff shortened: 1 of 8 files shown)


steveax:
👍

PR #362, opened 2016-10-07
Fix 64-bit puppet agent installation on non-English Windows

smcclellan:
When using a non-English version of the Windows installer, the architecture
of the machine was not being properly read. The install.ps1.erb script was
looking for "64-bit", but the "bit" part of that gets translated.

To solve this issue, the script should match on the string "64" instead.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-941


steveax:
👍

PR #364, opened 2016-10-12
Add broker for Puppet 4

smcclellan:
Now that Puppet 3 is approaching EOL, we should include a stock broker for
Puppet 4. Since the ideal name for the broker is `puppet`, the right approach
is to rename existing `puppet` broker to `legacy-puppet` and add a new `puppet`
broker for the new AIO broker.

If the user wants to change a policy to point to the old `puppet` broker, the
new `update-policy-broker` command will facilitate this. That will allow a
seamless migration.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-934

### DIFF
--- a/brokers/legacy-puppet.broker/install.erb
+++ b/brokers/legacy-puppet.broker/install.erb
@@ -60,9 +60,9 @@ case "${flavour}" in
     centos|redhat*)
         release="$(lsb_release -r | cut -f2 | cut -d. -f1)"
         case "${release}" in
-            5) url="https://yum.puppetlabs.com/puppetlabs-release-pc1-el-5.noarch.rpm" ;;
-            6) url="https://yum.puppetlabs.com/puppetlabs-release-pc1-el-6.noarch.rpm" ;;
-            7) url="https://yum.puppetlabs.com/puppetlabs-release-pc1-el-7.noarch.rpm" ;;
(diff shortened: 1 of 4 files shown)


brokers/legacy-puppet.broker/install.erb
```diff
@@ -0,0 +1,173 @@
...
+# Now we have that, we can install our platform specific package to define
+# the repositories.  Thankfully this is also pretty damn simple to get right.
+# In this case we don't skip installation even if the package was already
+# present, since the tool should handle that, and it may gather updates as a
+# consequence of taking the action.
+flavour="$(lsb_release -i | cut -f 2 | tr '[:upper:]' '[:lower:]')"
+case "${flavour}" in
+    centos|redhat*)
+        release="$(lsb_release -r | cut -f2 | cut -d. -f1)"
+        # @todo danielp 2013-09-30: this should be updated to reflect the
```
steveax: @smcclellan, seems like this comment is no longer valid? Can it be removed?

steveax:
Other than that old comment, LGTM!

brokers/legacy-puppet.broker/install.erb
smcclellan: Updated.

steveax:
image

LGTM!

PR #365, opened 2016-10-20
Fix handling of `hw_info`

smcclellan:
Two features related to `hw_info`, bundled into one PR to avoid a rebase race condition.

* Allow `mac` in `set-node-hw-info`

One confusing aspect about how Razor identifies nodes is that the mac
addresses are passed in as `net0`, `net1`, etc., but are represented in Razor
as `mac`, an array of all of these. To mitigate that confusion, the
`set-node-hw-info` command will now accept the `mac` attribute directly.

Additionally, when only `net0` was supplied, it was not being properly
converted into `mac` before the validation checks. This meant an error was
thrown saying that `mac` is a required attribute.

Now, the user can supply `netX`, `mac`, or both methods.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-939

* Store all `hw_info` when nodes are initially registered
Whenever a node is initially added to the system, the only `hw_info` values
that are recorded are the values that are used to match the node. This can be
a problem if the value used to match nodes changes, because there will be no
previous value to compare to in order to find the same node in the system.

To solve this, we will store all hardware info values that are passed by iPXE.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-958

### DIFF
--- a/lib/razor/command/set_node_hw_info.rb
+++ b/lib/razor/command/set_node_hw_info.rb
@@ -43,7 +43,7 @@ Update `node172` with new hardware information:
         --hw-info asset=Asset-1234567890 \\
         --hw-info uuid="Not Settable"
 
-With positional arguments, this can be shortened::
+With positional arguments, this can be shortened:
 
     razor set-node-hw-info node172 \\
         --hw-info mac=78:31:c1:be:c8:00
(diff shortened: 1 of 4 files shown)


steveax:
Merged manually so as not to squish.

PR #366, opened 2016-11-07
(maint) pin json-schema version

steveax:
This commit pins json-schema to v2.6.2 as later versions pull
in (through addressable) v2.0.4 of public_suffix which is
incompatible with ruby < v2.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,7 @@ group :test do
   gem 'simplecov'
   gem 'fabrication', '~> 2.7.2'
   gem 'faker', '~> 1.2.0'
-  gem 'json-schema', '~> 2.0'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -48,7 +48,7 @@ group :test do
   gem 'simplecov'
   gem 'fabrication', '~> 2.7.2'
   gem 'faker', '~> 1.2.0'
-  gem 'json-schema', '~> 2.0'
+  gem 'json-schema', '2.6.2'
```
smcclellan: It would be helpful to add a comment here about why we are pinning this specific version, noting that when ruby is upgraded, the version pin can be removed.

steveax:
@smcclellan updated

PR #367, opened 2016-11-08
(maint) Fix unparseable JSON input regression

smcclellan:
In a recent update to json-schema, inputs to the /api endpoints that are not a
JSON object or array return an error message besides "unparseable json". This
preserves previous behavior and also updates the Gemfile.lock.


steveax:
👍

PR #368, opened 2016-12-11
Add SLES task

smcclellan:
This adds a task to install SUSE Linux Enterprise Server version 12. It
conforms to the existing standard for default node metadata values, including
'hostname', 'timezone', and 'root_password'.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-966


steveax:
👍

smcclellan:
I have confirmed that this task also works for SLES 11 SP3. I've restructured the task to indicate it is for 11 (`sles/11` task) and also 12 (`sles/12` task).

PR #369, opened 2017-01-11
Add Windows Server 2016 task

smcclellan:
Razor should ship with an installer for Windows Server 2016. This adds a task
for that.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-975


steveax:
LGTM!

PR #370, opened 2017-01-11
Use single source Windows agent installer for puppet-pe broker

smcclellan:
The PE puppet agent installer for Windows is now installed via a Powershell
script downloaded via the puppet master, similar to how the Linux installer
works. To make the Windows broker standardized, this downloads and runs the
install.ps1 file from the puppet master.

The existing `windows_agent_download_url` and `version` metadata are no
longer used. To facilitate backwards compatibility, these will remain in the
configuration, but have been marked as deprecated in their description.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-974


steveax:
LGTM!

PR #371, opened 2017-02-06
Sanboot

ChrisPortman:
Boot Local Task Sanboot by Tag

This adds the option to detect machines that should use the sanboot iPXE
command through the use of a 'sanboot' tag.

Tags allow a policy based approach to selecting the hardware platforms
that require this boot method.  For example, creating a sanboot tag that applies
to all HP m300 moonshot cartridges which require the sanboot
method through the use of the 'productname' fact.


ChrisPortman:
Hmmm. seems to have added REALLY old commits to the PR that should not be there.

PR #373, opened 2017-02-17
Fix command execution when `auth.allow_localhost` is used

smcclellan:
Previously whenever a command execution was attempted, the
`auth.allow_localhost` functionality would cause a failure, meaning
non-authenticated localhost commands to return a 500 error. This commit results
in the intended behavior for commands as well as collections.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-994


steveax:
image
LGTM!

PR #375, opened 2017-03-21
Link to newest prebuilt microkernel (008)

darthmdh:
Updating the link to the prebuilt archive, 008 was released on 2-2-2017 and contains important fixes for booting on modern hardware.


smcclellan:
Thanks for pointing this out! I just updated these docs to use a link which will always direct the user to the latest version: https://github.com/puppetlabs/razor-server/commit/23d2164e8b6fbe6d34c85413cfae4763e7bbbf14

PR #378, opened 2017-09-26
(maint) Only require one of config and config defaults

smcclellan:
In order to accommodate both the PE and FOSS versions of razor-server, the
logic for finding the config file should require that only one of the config
and config defaults files exists.

### DIFF
--- a/lib/razor/config.rb
+++ b/lib/razor/config.rb
@@ -24,42 +24,53 @@ module Razor
 
     def initialize(env, fname = nil, defaults_file = nil)
       @values = {}
-      # Load defaults first.
+      # Find the defaults file.
+      # Priority: 1) environment variable
+      # 2) default path in /opt
+      # 3) root directory
       defaults_file ||= ENV["RAZOR_CONFIG_DEFAULTS"] ||
           (File.file?('/opt/puppetlabs/server/apps/razor-server/config-defaults.yaml') and '/opt/puppetlabs/server/apps/razor-server/config-defaults.yaml') ||
           File::join(File::dirname(__FILE__), '..', '..', 'config-defaults.yaml')
-      begin
+
+      # Find the user-maintained config.yaml
+      # Priority: 1) environment variable
+      # 2) default path in /etc
+      # 3) root directory
+      fname ||= ENV["RAZOR_CONFIG"] ||
+          (File.file?('/etc/puppetlabs/razor-server/config.yaml') and '/etc/puppetlabs/razor-server/config.yaml') ||
+          File::join(File::dirname(__FILE__), '..', '..', 'config.yaml')
+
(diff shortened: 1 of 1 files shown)


lib/razor/config.rb
```diff
@@ -24,42 +24,53 @@ class Config
...
       defaults_file ||= ENV["RAZOR_CONFIG_DEFAULTS"] ||
           (File.file?('/opt/puppetlabs/server/apps/razor-server/config-defaults.yaml') and '/opt/puppetlabs/server/apps/razor-server/config-defaults.yaml') ||
           File::join(File::dirname(__FILE__), '..', '..', 'config-defaults.yaml')
-      begin
+
+      # Find the user-maintained config.yaml
+      # Priority: 1) environment variable
+      # 2) default path in /etc
+      # 3) root directory
+      fname ||= ENV["RAZOR_CONFIG"] ||
```
steveax: I think it'd be clearer if this had a less generic name, maybe: `config_defaults_file` and config_user_file`?

lib/razor/config.rb
```diff
@@ -24,42 +24,53 @@ class Config
...
         yaml = File::open(defaults_file, "r") { |fp| YAML::load(fp) } || {}
 
         @values.merge!(yaml["all"] || {})
         @values.merge!(yaml[Razor.env] || {})
-      rescue Errno::ENOENT
-        # The defaults file does not exist. This is okay.
-      rescue Errno::EACCES
+      elsif File.exists?(defaults_file)
         raise InvalidConfigurationError,
               _("The configuration defaults file %{filename} is not readable") % {filename: defaults_file}
```
steveax: Is this testing if it's present, or readable?

lib/razor/config.rb
```diff
@@ -24,42 +24,53 @@ class Config
...
-      rescue Errno::ENOENT
-        raise InvalidConfigurationError,
-          _("The configuration file %{filename} does not exist") % {filename: fname}
-      rescue Errno::EACCES
+        @values.merge!(yaml["all"] || {})
+        @values.merge!(yaml[Razor.env] || {})
+      elsif File.exists?(fname)
         raise InvalidConfigurationError,
-          _("The configuration file %{filename} is not readable") % {filename: fname}
+              _("The configuration file %{filename} is not readable") % {filename: fname}
```
steveax: "is present but not readable"?

lib/razor/config.rb
steveax: ahh, I see, maybe then: "is present but not readable"?

steveax:
Aside from a few nits, 👍

smcclellan:
Already fixed via 1355bc13881b45f93e53584cfc38a949e2349966.

PR #379, opened 2017-09-26
Add `has_macaddress_like` operator

smcclellan:
In order to fuzzily match macaddresses in tags, this adds a
`has_macaddress_like` operator, which accepts a variable list of regex strings.

Fixes https://github.com/puppetlabs/razor-server/RAZOR-783


steveax:
👍

PR #380, opened 2017-09-27
(RE-8447) Migrate pe-razor to vanagon

mwaggett:
This commit adds files from pe-razor-server so that we can deprecate pe-razor-server and only maintain this repo. razor-server will (eventually) promote into PE the same way that puppet-agent does.


mwaggett:
Most of this was adding the `acceptance/` and `integration/` dirs for PE.

acceptance/lib/helper.rb
```diff
@@ -0,0 +1,170 @@
+require 'beaker/dsl/install_utils'
+
+module RazorExtensions
+  def self.initialize_config(options)
+
+    install_type = get_option_value(options[:pe_razor_server_install_type],
+                                    [:package], "Install Type",
+                                    "PE_RAZOR_SERVER_INSTALL_TYPE", :package)
```
melissa: you're probably going to want to update all the references to pe-razor-server to razor-server, since we hopefully won't be building pe-razor-server packages anymore

mwaggett: ooh yeah, good call

mwaggett: @smcclellan I removed obvious references to the pe-razor-server service/package in these tests, but I'm unsure about whether references like this (`:pe_razor_server_install_type`, e.g.) need to/should be changed, since these tests are still for a PE context. Thoughts?

PR #381, opened 2017-09-27
(maint) update jruby to 1.7.26

steveax:
jruby-1.7.19 is no longer on the travis ci list of rubies. This
commit bumps to jruby-1.7.26.


steveax:
Need to do some investigation about how this might affect acceptance tests.

smcclellan:
Next jruby bump will be to 9.1.5.0 to follow torquebox's bundled version.

PR #382, opened 2017-11-29
(RAZOR-1022) Fix `repo_file` task function for repos with remote source

smcclellan:
The `repo_file` function, accessible by tasks, currently only looks for local
files in the repo directory. This works fine when the repo is created with the
`iso_url` argument, but not when it's using a remote source via the `url`
argument.

The fix is for this function is to download from a remote source when `url` is
set, otherwise use the current behavior of using a local source's contents.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-1022

PR #383, opened 2017-11-29
(RAZOR-967) Add ESXi 6 support

smcclellan:
This adds stock task support for ESXi 6. This is a nominal change which will
allow future expansion as we support more versions. Following the standard of
other tasks, the `vmware_esxi` task will always reference the latest version of
the OS that we support. This commit also adds two new tasks: `vmware_esxi/5.5`
and `vmware_esxi/6`. These two tasks are functionally identical, but if we add
a later task that requires changes to the task's files, the 5.5 and 6 tasks will
remain functionally identical to how they are now.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-967


smcclellan:
Note: This is blocked by https://github.com/puppetlabs/razor-server/pull/382.

steveax:
👍

PR #384, opened 2017-11-30
(RAZOR-854) Add Ubuntu Xenial task

smcclellan:
This adds a task for Ubuntu Xenial. Since that is the latest version of Ubuntu
that we support, that also means the generic `ubundu` task should point to this.
An `ubuntu/xenial` task now exists, which will work for Xenial even when new
versions are added.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-854

### DIFF
--- a/tasks/ubuntu/xenial.task/metadata.yaml
+++ b/tasks/ubuntu/xenial.task/metadata.yaml
@@ -1,0 +1,6 @@
+---
+
+os_version: xenial
+label: Ubuntu Xenial 14.04
+description: Ubuntu Xenial Installer
+base: ubuntu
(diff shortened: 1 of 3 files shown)


steveax:
👍

tasks/ubuntu/xenial.task/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+---
+
+os_version: xenial
+label: Ubuntu Xenial 14.04
```
jharbott: This should be 16.04

steveax: Indeed! Good catch.

smcclellan: I fixed this [in a follow-up commit](https://github.com/puppetlabs/razor-server/commit/6754ab679839e936dd4072ad1286cae6bc6ac086).

PR #386, opened 2017-12-08
(maint) Add gem version bindings since many gems are dropping 1.9.3

underscorgan:
This still uses a 1.9.3 jruby. Since many gems are moving to requiring
more recent versions of ruby, or dependencies requiring more recent
versions of ruby, we need to make sure we're binding to a version that
works with our dependencies.


underscorgan:
@smcclellan updated! and testing at least an el7 build it looks to be working 🎉

PR #387, opened 2017-12-11
Add UEFI Support

smcclellan:
This PR does a few things: 

- Allows `hw_info` not used to match to remain with the node. In the past, these would be removed.
- Adds a `fact_boot_type` value to `hw_info` that can be used in tasks.
- Changes the partitioning info in the Windows unattended.xml files to handle UEFI if the `fact_boot_type` value is set to `efi`.

This fixes https://tickets.puppetlabs.com/browse/RAZOR-1025.


steveax:
👍

PR #388, opened 2017-12-11
Update unattended.xml.erb

kreeuwijk:
this adds support for UEFI based on a new fact_platform item from hw_info


smcclellan:
Tracking this change in https://github.com/puppetlabs/razor-server/pull/387.

PR #389, opened 2017-12-14
Modifications to support Windows ADK 10

smcclellan:
ADK 10 requires that the WinPE image have a specific name of `boot.wim`.
This alters the output from the build-razor-winpe.ps1 to match that
convention instead of the previous `razor-winpe.wim`. Renaming here
should save the user a step when the WIM is copied to the Razor server.

Additionally, this adds fake credentials to the SMB connection in the
Windows task. This is required for ADK 10 to mount an anonymous
fileshare.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-1024


steveax:
👍

PR #391, opened 2018-01-31
Update to JRuby 9000 and Torquebox 3.2.0

smcclellan:
To stay ahead of updates to our dependencies and ensure our
gems get proper security fixes, we should move to JRuby 9.1.5.0. This
also allows us to update to the latest version of Torquebox: 3.2.0.

This commit also addresses a few things that have changed in our
dependencies.

On the sequel gem side, serialization now happens prior to validation.
That has two implications: we need to re-serialize objects if they
change in their validation, and serialization needs to be robust enough
that it will not fail prior to the object being validated.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-973


McdonaldSeanp:
🙈 

👍

steveax:
image
LGTM!

johnf:
When should this make it into the released deb packages via https://apt.puppetlabs.com/puppetlabs-release-pc1-xenial.deb?

Currently with older ruby TLS1.2 isn't supported which means https simply fails on some https sites. e.g. mirror.aarnet.edu.au

PR #392, opened 2018-02-06
Exclude `fact_boot_type` from hw_info matching

smcclellan:
When nodes are matched, every item in `hw_info` is used and must be
distinct for nodes across a deployment. Since the boot method of the
node should ideally be passed in via iPXE, we need a special case which
persists this value in the database but doesn't use it for matching,
since it's not unique.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-1045


steveax:
👍

PR #394, opened 2018-02-16
Skip certain tests if the user is running as root

ekinanp:
Previously, a test in spec/broker_type_spec.rb and another test in
spec/data/repo_spec.rb set the permissions of a file to "0000" with
the intent that the codebase would throw an exception if that file
was not readable. When running as root, however, everything is
readable so these tests failed. This commit skips these tests when
running as root, because they don't make sense for that specific
case.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-1050

### DIFF
--- a/spec/broker_type_spec.rb
+++ b/spec/broker_type_spec.rb
@@ -137,6 +137,9 @@ describe Razor::BrokerType do
       end
 
       it "should raise if the install template is present but unreadable" do
+        pending("Skipping this test because process is running with " +
+                "root user privileges and root can read everything") if Process.uid == 0
+
         broker = {'test' => {'install.erb' => "# no real content here\n"}}
         with_brokers_in(paths.first => broker) do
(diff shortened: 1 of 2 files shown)


spec/broker_type_spec.rb
```diff
@@ -137,6 +137,8 @@ def with_brokers_in(brokers)
       end
 
       it "should raise if the install template is present but unreadable" do
+        pending("Skipping this test because process is running with root user privileges and root can read everything") if Process.uid == 0
```
smcclellan: Can we break this line so the line length is more standard with the rest of the file? e.g.
```
  pending("Skipping this test because process is running with " +
          "root user privileges and root can read everything") if Process.uid == 0
```

smcclellan:
@ekinanp It's also more standard in the Razor codebase to add `Fixes <full ticket URL>` at the end of a commit description instead of putting the ticket number in the commit summary in parens.

PR #395, opened 2018-02-16
Pass-thru if stdin.write fails with Errno::EPIPE in exec_script

ekinanp:
Previously, the tests in spec/data/hook_spec.rb would transiently
fail at the stdin.write step in exec_script due to an Errno::EPIPE
exception. Errno::EPIPE is thrown because of a race condition: the
process running the script exited before stdin.write was called. This
happened because the test scripts finished very quickly. From a user's
perspective, this would mean that the exec_script task has a chance
of failure if the executed script finishes quickly, which is undesirable.

This commit fixes things so that when stdin.write fails with an Errno::EPIPE
exception, the code just passes through. This should be OK, because if something
goes wrong with the process itself, that should be captured in the process' success
status queried at the end of exec_script.

The code also cleans up exec_script to improve its readability.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-1051

### DIFF
--- a/lib/razor/data/hook.rb
+++ b/lib/razor/data/hook.rb
@@ -318,37 +318,38 @@ class Razor::Data::Hook < Sequel::Model
       # Run the file from the hook's directory so that relative paths work.
       hook_dir = File.expand_path('..', script)
       stdin, stdout, stderr, wait_thr = Bundler.with_clean_env do
-        begin
-          old = Dir.pwd
-          extra_path = Razor.config['hook_execution_path']
-          ENV['PATH'] = "#{extra_path}:#{ENV['PATH']}" if extra_path
-          Dir.chdir(hook_dir)
+        extra_path = Razor.config['hook_execution_path']
+        ENV['PATH'] = "#{extra_path}:#{ENV['PATH']}" if extra_path
+        Dir.chdir(hook_dir) do
           Open3.popen3(script.to_s)
-        ensure
-          Dir.chdir(old)
         end
       end
-      stdin.write(args) if args
+
       begin
-        stdin.close unless stdin.closed?
-      rescue Errno::EPIPE, Errno::EBADF, IOError
+        stdin.write(args) if args
+      rescue Errno::EPIPE
         # Do nothing; this means the hook did not read stdin or that stdin
         # closed already.
       end
+
+      # This is necessary for the join call below, as otherwise the thread will not
+      # finish
+      stdin.close
       wait_thr.join
-      # Prefer nil over an empty string.
+
       output = stdout.readlines.join
       error = stderr.readlines.join
       process = wait_thr.value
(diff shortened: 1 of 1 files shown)


lib/razor/data/hook.rb
```diff
@@ -318,37 +318,38 @@ def exec_script(script, args)
...
       hook_dir = File.expand_path('..', script)
       stdin, stdout, stderr, wait_thr = Bundler.with_clean_env do
-        begin
-          old = Dir.pwd
-          extra_path = Razor.config['hook_execution_path']
-          ENV['PATH'] = "#{extra_path}:#{ENV['PATH']}" if extra_path
-          Dir.chdir(hook_dir)
+        extra_path = Razor.config['hook_execution_path']
+        ENV['PATH'] = "#{extra_path}:#{ENV['PATH']}" if extra_path
+        Dir.chdir(hook_dir) do
```
ekinanp: Looking at JRuby's source (and chdir's specs), if chdir is passed a block, it will ensure that the old directory is restored after the code executes even if an exception is thrown. Here is the relevant part of the JRuby source:

https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/RubyDir.java#L310-L319

Also the docs on chdir: http://ruby-doc.org/core-2.2.3/Dir.html#method-c-chdir

lib/razor/data/hook.rb
```diff
@@ -318,37 +318,38 @@ def exec_script(script, args)
...
-      # Prefer nil over an empty string.
+
       output = stdout.readlines.join
       error = stderr.readlines.join
       process = wait_thr.value
+      # Prefer nil over an empty string.
       [process.exitstatus, process.success?, output.empty? ? nil : output,
        error.empty? ? nil : error]
-    rescue IOError => e
+    rescue IOError, SystemCallError => e
```
ekinanp: Looking at the docs, http://ruby-doc.org/core-2.2.0/IOError.html indicates that SystemCallErrors are not captured in IOError, despite some IO operations having the possibility of raising them.

lib/razor/data/hook.rb
```diff
@@ -318,37 +318,38 @@ def exec_script(script, args)
...
       # Error running the script. Catch the message and continue.
       [1, false, nil, e.message]
     ensure
-      stdin.close unless !stdin or stdin.closed?
-      stdout.close unless !stdout or stdout.closed?
-      stderr.close unless !stderr or stderr.closed?
+      stdin.close
```
ekinanp: This is safe, popen3 always returns something (based on:
  https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/open3.rb#L102 and https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/open3.rb#L209)

PR #396, opened 2018-02-16
(maint) Remove old packaging

smcclellan:
Now that Razor is a vanagon-based project, we don't need these packaging
tasks anymore.


ext/project_data.yaml
```diff
@@ -1,51 +0,0 @@
...
-  choose which preconfigured model to apply to a new node, which will begin to
-  follow the model's directions, giving feedback to Razor as it completes various
-  steps. Models can include steps for handoff to a DevOps system such as Puppet
-  or to any other system capable of controlling the node (such as a vCenter
-  server taking possession of ESX systems).
-# automatically burn our version into the generated source
-version_file: 'lib/razor/version.rb'
-update_version_file: true
-# files and gem_files are space separated lists (or arrays)
-files:
```
smcclellan: Does this list actually exist somewhere in razor-vanagon, or should it?

underscorgan: I believe this list is only needed for old-fashioned building with the packaging repo

PR #397, opened 2018-02-20
(RAZOR-1057) Port over pe-razor-server specific items to razor-server

ekinanp:
Pretty much ports over whatever's discussed in https://docs.google.com/document/d/1vFF7g-Dbjr5MJvu3PEsU4_u0cPIrzHDWY1BfXoOqyrk/edit to razor-server. Note that I did not nix the razor-server init script b/c that is work that can be done in a separate ticket (once we get the module setup, I think).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,13 @@
 require 'rake'
-require 'yaml'
+require 'pathname'
+JRUBY_BIN  = "/usr/local/share/pl-jruby/bin/jruby -S"
+RAKE_ROOT = File.expand_path(File.dirname(__FILE__))
+
+begin
+  require 'torquebox-rake-support'
+rescue LoadError
+  STDERR.puts "Unable to load 'torquebox-rake-support'. Some rake tasks may be unavailable without this library."
+end
 
 task :default do
   system("rake -T")
@@ -46,21 +54,19 @@ namespace :db do
   task :reset, [:env] => [:nuke, :migrate]
 end
 
-namespace :spec do
-  begin
+if defined?(RSpec::Core::RakeTask)
+  namespace :spec do
     require 'rspec/core'
     require 'rspec/core/rake_task'
-
+  
     task :reset_tests do
       Rake::Task['db:reset'].invoke("test")
     end
-
+  
     desc "Run all specs"
     RSpec::Core::RakeTask.new(:all => :reset_tests) do |t|
       t.pattern = 'spec/**/*_spec.rb'
     end
-  rescue LoadError
-    # ignore
   end
 end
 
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -1,5 +1,13 @@
 require 'rake'
-require 'yaml'
+require 'pathname'
```
ekinanp: This excerpt is from pe-razor-server: https://github.com/puppetlabs/pe-razor-server/blob/2018.1.x/Rakefile#L3-L10

Rakefile
```diff
@@ -46,21 +54,19 @@ namespace :db do
   task :reset, [:env] => [:nuke, :migrate]
 end
 
-namespace :spec do
-  begin
+if defined?(RSpec::Core::RakeTask)
```
ekinanp: pe-razor-server: https://github.com/puppetlabs/pe-razor-server/blob/2018.1.x/Rakefile#L58-L72

Rakefile
```diff
@@ -79,93 +85,29 @@ task :console do
   IRB.start
 end
 
-desc "Build an archive"
```
ekinanp: The :archive task is still there. I wonder if it can be removed since it seems like packaging?

bin/razor-admin-wrapper
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
ekinanp: As discussed in the doc, this can be safely ported over.

lib/razor/config.rb
```diff
@@ -54,19 +54,23 @@ def initialize(env, fname = nil, defaults_file = nil)
...
-        @values.merge!(yaml["all"] || {})
-        @values.merge!(yaml[Razor.env] || {})
-      rescue Errno::ENOENT
-        # This is only an error if we don't have any config values yet.
-        raise InvalidConfigurationError,
-          _("The configuration file %{filename} does not exist") % {filename: fname} if @values.empty?
-      rescue Errno::EACCES
-        raise InvalidConfigurationError,
-          _("The configuration file %{filename} is not readable") % {filename: fname}
+      if File.readable?(fname)
```
ekinanp: Main change in https://github.com/puppetlabs/pe-razor-server/commit/bb315d013b113dc480bc101e89b060abfc6e7115 for this file.

lib/razor/config.rb
```diff
@@ -54,19 +54,23 @@ def initialize(env, fname = nil, defaults_file = nil)
...
+        begin
+          yaml = File::open(fname, "r") { |fp| YAML::load(fp) } || {}
+
+          @values.merge!(yaml["all"] || {})
+          @values.merge!(yaml[Razor.env] || {})
+        rescue Errno::ENOENT
+          # This is only an error if we don't have any config values yet.
+          raise InvalidConfigurationError,
+            _("The configuration file %{filename} does not exist") % {filename: fname} if @values.empty?
```
ekinanp: I wonder if this is something we want to raise outside of the if? As in, do we need to have a configuration before returning from the initialize method in the config.rb class or only when we know that fname exits?

Rakefile
smcclellan: I've been using this `archive` task historically to post a zip to AWS for direct download, but that's apparently used very infrequently. I am 👍to deleting this.

lib/razor/config.rb
smcclellan: Let's just nix the `readable?` if block here. The `if @values.empty?` bit catches the main concern here.

ekinanp:
Updated the PR to remove the :archive task and also to remove the readable? check in lib/razor/config.rb

lib/razor/config.rb
```diff
@@ -64,9 +64,10 @@ def initialize(env, fname = nil, defaults_file = nil)
         raise InvalidConfigurationError,
           _("The configuration file %{filename} does not exist") % {filename: fname} if @values.empty?
       rescue Errno::EACCES
-        raise InvalidConfigurationError,
-          _("The configuration file %{filename} is not readable") % {filename: fname}
+        # The given file is not readable. This is okay.
```
smcclellan: I think this should actually throw an error. Anytime a referenced file isn't readable via permissions, I'd expect this to throw.

ekinanp: This achieves the equivalent effect of the `readable?` check -- if the file is not readable in the original code, it is a noop just like it is here.

smcclellan: Not exactly, since `readable?` also returns `true` if the file does not exist. If the file exists, and can't be read, my intuition tells me that should be an error.

ekinanp: Oh ok, if that's what we want to do then I updated the PR to reflect that.

lib/razor/config.rb
```diff
@@ -65,8 +65,10 @@ def initialize(env, fname = nil, defaults_file = nil)
           _("The configuration file %{filename} does not exist") % {filename: fname} if @values.empty?
       rescue Errno::EACCES
         raise InvalidConfigurationError,
-          _("The configuration file %{filename} is not readable") % {filename: fname}
+          _("The configuration defaults file %{filename} is not readable") % {filename: fname}
```
smcclellan: I think this is a copy/paste error?

ekinanp: Yup, updated the PR to fix that.

PR #398, opened 2018-02-27
hostname should not be fqdn

chalb:
- at this point the hostname shouldn't be the fqdn


smcclellan:
This is a change we are planning to make, but we need a bit more nuance, described here:
https://tickets.puppetlabs.com/browse/RAZOR-665?focusedCommentId=531286&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-531286

Thanks for the PR, as it validates that this is an issue that affects more people than we might've thought! I'll have a separate PR up soon to address this ticket. In the meantime, feel free to override the `set_hostname.erb` file in your local install.

PR #400, opened 2018-04-02
Remove EOL Ubuntu Tasks

smcclellan:
Ubuntu Lucid and Ubuntu Precise are both past their End-of-life. Since
we don't want to support platforms that aren't supported upstream,
these should be removed from our stock tasks.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-800


NEWS.md
```diff
@@ -14,7 +22,7 @@
 + NEW: Added `repo_file_contents` method for use in templates that
   returns the contents of a file in the repo.
 + NEW: Added `repo_file?` method for use in task templates that returns
-  a path to a file in a repo, if it exists. This will replace the
+  a path to a fi`le in a repo, if it exists. This will replace the
```
branan: Looks like a typo snuck in here

PR #401, opened 2018-04-10
(RAZOR-1065) Move service user declaration to environment variable

mwaggett:
This commit adds a `RAZOR_USER` environment variable and sets it to 'razor'.
When building packages for PE, this variable will be changed to 'pe-razor'.


smcclellan:
I think the commit summary should be something more specific, like "Moving user declaration into environment variable"?

mwaggett:
sure, I can do that.

PR #405, opened 2018-04-17
Add a 'depth' parameter to the /api/collections/<type> endpoint

ekinanp:
Previously, the /api/collections/<type> endpoint would return a list
of object references to the items in the collection. For people who
wanted to see the complete details of each item in the collection,
they would have needed to iterate over the item references and make
a total of N requests to their corresponding API endpoints, where N
is the size of the collection. This is annoying when writing scripts,
and it does not scale well for large collections.

This commit introduces an optional depth parameter that can be one
of two possible values:
  (1) 0 -- Returns a list of object references for the items
  in the collection, which is the default behavior.

  (2) 1 -- Returns a detailed list of the items in the collection,
  meaning that for each entry <item> in the collection, the response body
  of the endpoint /api/collections/<type>/<item> is returned.

The default depth is 0.

Linked to https://tickets.puppetlabs.com/browse/RAZOR-433


ekinanp:
The bulk of the changes here are to the unit tests, the feature change is small.

It would be nice to DRY up the depth parameter tests, which I can do if that is desired (e.g. have a function `run_depth_parameter_tests(col_type, expected_size, item_schema)`) and use that.

app.rb
```diff
@@ -638,9 +638,31 @@ def render_template(name)
...
+  ].map do |coll|
+    # All collection elements save for :config will have a depth parameter
+    # which has the following semantics:
+    #   depth = 0 will return the list of items in the collection
+    #
+    #   depth = 1 will return a detailed list of items in the collection,
+    #   i.e. for each item, it will return whatever collection/<coll>/<item_id>
+    #   returns.
+    next coll if coll == :config
+    depth_param = { 'depth' => { "type" => "number" } }
```
smcclellan: I'm not sure the algorithm here is necessary. I'd lean toward something like this:
```
start = {'start' => {'type' => 'number'}}
limit = ...
depth = ...
COLLECTIONS = [
  [:brokers, depth],
  [:repos, depth],
  ...
  [:nodes, start, limit, depth]
  ...
]
```
That to me makes it simpler regarding what is available for each collection. Variable names for `start`, `limit`, `depth` could be better.

spec/app/api_spec.rb
```diff
@@ -22,43 +23,588 @@ def validate!(schema, json)
...
-    'title'    => "Object Reference Collection JSON Schema",
+  ObjectRefItemSchema = {
     'type'     => 'object',
     'additionalProperties' => false,
     'properties' => {
       "spec" => {
         'type'    => 'string',
         'pattern' => '^https?://'
       },
-      "items" => {
```
smcclellan: Is `items` going away here? I thought this wouldn't be an API change at all, just an addition.

ekinanp: No, that is the schema for an item in a collection that's an object ref. See the collection_schema routine later below api_spec.

Otherwise if you look at collection_view in lib/razor/view.rb, the API's unchanged.

ekinanp:
@smcclellan Updated the PR with the following changes:

(1) The error response for an invalid depth parameter now includes the value passed-in by the user.

(2) The COLLECTIONS constant + the location it was used was modified to make it easier to understand.

(3) The schemas were moved to a separate file, spec/lib/razor/spec/collection_schemas.rb. I chose to add an extra spec namespace since `Razor::CollectionSchemas` seems like something would make sense in an application, whereas `Razor::Spec::CollectionSchemas` makes it clearer that those schemas are for the spec tests.

(4) The depth parameter tests were refactored to use a shared example. This DRY'd up the api_spec code quite a bit, so I think we're OK w/ leaving things in that file now. Note that with (3) and (4), api_spec has fewer lines of code than it previously did.

lib/razor/view.rb
```diff
@@ -219,9 +219,18 @@ def collection_view(cursor, name, args = {})
...
-        map {|t| view_object_reference(t)}.
-        select {|o| check_permissions!("#{perm}:#{o[:name]}") rescue nil }
+
+      # Check the depth parameter, then compute the "items" hash
+      valid_depths = [0, 1].map(&:to_s)
+      depth = args[:depth] || "0"
+      unless valid_depths.include?(depth)
+        error 400, :error => _("'#{depth}' is not a valid value for the depth parameter. Valid values are 0 or 1")
+      end
+      hash_fn = depth == "0" ? :view_object_reference : args[:hash_fn]
```
smcclellan: Can you add `depth` and `args[:hash_fn]` to the documentation for this method?

spec/spec_helper.rb
```diff
@@ -140,6 +140,18 @@ def run_message(message)
...
 
+# Include the Razor::Data item schemas + related schema helpers for the
+# api spec tests, specifically the /api/collection endpoint spec tests.
+# Note that we chose to put these in a separate file, because they take
+# up a lot of space without adding much additional clarity.
+require_relative 'lib/razor/spec/collection_schemas'
+RSpec.configure do |c|
+  # extend is necessary to use the helpers outside of an "it" block;
+  # include is to use the helpers inside an "it" block.
+  c.extend Razor::Spec::CollectionSchemas, api_spec: true
```
smcclellan: What purpose does this `api_spec` serve?

ekinanp: That's to limit including those helpers to only the api_spec tests. If you look at api_spec, you'll see the `api_spec: true` annotation next to the describe. Otherwise without it, those helpers will be included in all tests.

Will add a clarifying comment above.

smcclellan: Are you concerned about a performance slowdown to having those helpers available in other tests?

ekinanp: No, I just don't see a need to include them in other tests when they're solely meant for api_spec.

doc/api.md
```diff
@@ -759,7 +759,7 @@ Along with the list of supported commands, a `GET /api` request returns a list
 of supported collections in the `collections` array. Each entry contains at
 minimum `url`, `spec`, and `name` keys, which correspond respectively to the
 endpoint through which the collection can be retrieved (via `GET`), the 'type'
-of collection, and a human-readable name for the collection.
+of collection, and a human-readable name for the collection.
```
smcclellan: It looks like your editor is keeping extra spaces in the file?

NEWS.md
```diff
@@ -2,6 +2,14 @@
 
 ## Next - Next
 
+### API changes
+
++ NEW: Added a `depth` parameter to the `/api/collections/<type>` endpoints,
+  save for when `<type> = config`, that has two possible values: `0` and `1`. `0`
+  will return a list of object references, while `1` will return the fully expanded
+  list of objects. `0` is the default behavior when the `depth` parameter is not
+  provided. See `doc/api.md` for more details
```
smcclellan: Missing a period here.

ChrisPortman:
Is it anticipated that at some point depth may take values other than 0 or 1?  If not, should depth be renamed to something that implies a boolean value?  Its made more ambiguous by the nature of the retrieved data structures being more than 1 level deep.  The name depth implies that it can do various levels of expansion.
I think my original PR used expand=true|false

smcclellan:
@ChrisPortman We debated several names for the property, and eventually decided it made more sense for it to describe how deep the results should go. It could feasibly go deeper in the future, for example (querying /api/collections/tags):
- depth = 0 includes references: /api/collections/tags
- depth = 1 includes full objects: /api/collections/tags/:name
- depth = 2 would include full objects for /api/collections/tags/:name/nodes

At this point it does seem unlikely that we'll use higher depths, but this seemed like an extensible and straightforward solution. Will it still handle the use case you're wanting it for?

Thanks also for the original pull request! And it's good to see you again. :)

PR #406, opened 2018-04-19
(RAZOR-1074) razor-service won't start on ubuntu-14.04

mwaggett:
This commit adds a check for the `runuser` command in the razor-server init
script. Ubuntu-14.04 does not have this command, so we fall back to using `su`
if `runuser` is not found.


mwaggett:
Builds with this change: http://builds.delivery.puppetlabs.net/razor-server/0c236050e31e888e93835555138310573d780866/

mwaggett:
Service starts successfully (tested on ubuntu-14.04, el-6, and el-7)!

PR #410, opened 2018-05-01
(maint) Consolidate config files

mwaggett:
This commit removes the razor-server.env file (used only by the systemd service
file) and moves all its variable assignments to the razor-server.sysconfig,
which is used by both the sysV init script and the systemd service file. This
way, it will be clearer and more consistent how the service receives arguments
and will allow these args to be managed by the puppetlabs-pe_razor module.


mwaggett:
should go in with https://github.com/puppetlabs/puppetlabs-pe_razor/pull/61 and https://github.com/puppetlabs/razor-vanagon/pull/38

mwaggett:
I'm closing this and reopening a new one because apparently I didn't branch before pushing this up and that was silly of me. 😝

PR #411, opened 2018-05-18
(maint) Consolidate config files

mwaggett:
This commit removes the razor-server.env file (used only by the systemd service
file) and moves all its variable assignments to the razor-server.sysconfig,
which is used by both the sysV init script and the systemd service file. This
way, it will be clearer and more consistent how the service receives arguments
and will allow these args to be managed by the puppetlabs-pe_razor module.


mwaggett:
should go in with puppetlabs/puppetlabs-pe_razor#61 and puppetlabs/razor-vanagon#38

mwaggett:
Tested foss installs on el-7-x86_64 and ubuntu-14.04-amd64.
Test PE upgrades from PE 2018.1.0 on el-6-x86_64 and el-7-x86_64.

PR #413, opened 2018-05-29
(PE-24400) Don't use variables in sysconfig

mwaggett:
This commit expands previously used environment variables into their intended
strings in the sysconfig. Since we use the sysconfig as an environment file for
the razor-server service and systemd only allows simple variable declaration,
we must declare the full strings in each case.

### DIFF
--- a/ext/razor-server.sysconfig
+++ b/ext/razor-server.sysconfig
@@ -1,3 +1,6 @@
+# This file is used as a systemd environment file, so we can't rely on variable
+# expansion. Instead, we must set each env var to its full string.
+#
 # The location of the razor config files
 RAZOR_CONFIG=/etc/puppetlabs/razor-server/config.yaml
 RAZOR_CONFIG_DEFAULTS=/opt/puppetlabs/server/apps/razor-server/config-defaults.yaml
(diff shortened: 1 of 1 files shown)


underscorgan:
Hm, trying to figure out what my thoughts are re: expanding out all the variables vs reverting https://github.com/puppetlabs/razor-server/pull/411/files

mwaggett:
@underscorgan I'm thinking that it will be a lot easier to manage all these variables if they're all in one place. When working on this before, it was pretty confusing to keep track of where everything was getting set and what was passing to what.

underscorgan:
okay. can you add a comment in the code for why you can't use variables so future folks looking for cleanup/optimization don't break this? It's not obvious from the code that this file is going to be a systemd environment file

ext/razor-server.sysconfig
```diff
@@ -1,3 +1,6 @@
+# This file is used as a systemd environment file, so we can't rely on variable
+# expansion. Instead, we must set each env var to its full string.
+#
```
mwaggett: @underscorgan added!

PR #414, opened 2018-05-29
(maint) Update NEWS.md for release

mwaggett:


NEWS.md
```diff
@@ -17,6 +17,8 @@
 + IMPROVEMENT: System config files for packaging various
   distributions are now consolidated.
 
+## 1.9.0 - Never Released
```
smcclellan: IMO this stub isn't needed. If we might be burning more tags in the future, let's just omit these stubs for them.

mwaggett:
@smcclellan updated!

PR #421, opened 2018-11-20
(maint) Add Docker support to run torquebox in container

smcclellan:
This adds two scripts:
- docker/bin/docker-start.sh: This script builds and starts two
  containers: one for postgres and one jruby container for razor-server.
  The repo_store_root will be mounted to the `repo/` directory, and the
  microkernel will be downloaded if it doesn't already exist.
- docker/bin/docker-stop.sh: This script stops and removes these
  containers.

This is useful for development in testing out API changes. Some future
use cases which require more work include:

- Testing tasks/provisioning workflows by PXE booting against this
  server. This requires an additional dnsmasq box to field DHCP.
- Run the container as a standalone server. Doing this properly
  requires swapping out the `postgres` container and using a
  persistent data store. There may be issues with queues not persisting
  as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ log/*
 /.project
 /.ruby-version
 /ext/packaging/
+/repo
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+FROM jruby:9.1.5.0-alpine
+
+# throw errors if Gemfile has been modified since Gemfile.lock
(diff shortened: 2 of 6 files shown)


smcclellan:
Should perhaps be combined with https://github.com/puppetlabs/razor-server/pull/428.

smcclellan:
Merged into https://github.com/puppetlabs/razor-server/pull/434.

PR #426, opened 2019-04-10
Add quotes around kickstart root password variable

runejuhl:
It turns out that these quotes are necessary to allow passwords with spaces, otherwise Razor install will fail with an argument error.

### DIFF
--- a/tasks/centos.task/kickstart.erb
+++ b/tasks/centos.task/kickstart.erb
@@ -7,7 +7,7 @@ url --url=<%= repo_url %>
 cmdline
 lang en_US.UTF-8
 keyboard us
-rootpw <%= node.metadata['root_password'] || node.root_password %>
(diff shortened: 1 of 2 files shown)


runejuhl:
FYI I'm not sure if similar quotes needs to be added to `debian.task` as well -- I'd guess that it should, but I haven't tested it.

smcclellan:
Thanks for the PR! I'll merge it as-is. It's generally best to have the config management solution (e.g. puppet) control the machine's password long-term, and just using a dummy password for setup in the kickstart/preseed. Either way, this may be helpful!

runejuhl:
Absolutely, that's how we use it too :smile: 

I just have a homebrew password generator script that I tend to use because it produces hard passwords that can be transferred fairly easily by voice, and had a few attempts at install before I realized that a password like `dumplings gb_ veriest / or` was the issue...

PR #428, opened 2019-04-17
(maint) Add rake task for Razor database in Docker

smcclellan:
This is a tool primarily for development, where the Razor
database can run as a docker container. If you run `rake docker:db`
then run spec tests, they will use the container. `rake docker:stop`
will stop the container.

Waiting to merge until PE version is in a release branch.


smcclellan:
Merged into https://github.com/puppetlabs/razor-server/pull/434.

PR #429, opened 2019-05-28
(RAZOR-1060) Add support for Ubuntu Bionic 18.04

speedofdark:
This commit adds support for Bionic and makes it the default install
for Ubuntu. Trusty and Xenial are also supported as discrete tasks.

Note that the 'live-server' image does not work; you need to use

http://cdimage.ubuntu.com/ubuntu/releases/18.04/release/ubuntu-18.04.2-server-amd64.iso

as the base image.


tasks/ubuntu/trusty.task/os_boot.erb
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
caseywilliams: Could these new templates go in a new `bionic.task` directory? Or is it necessary to update the trusty and xenial files here?

speedofdark:
Thats the way the file precedence works for razor configuration. We need
both the mainline task (ubuntu.task/) as well as the ancillary
ubuntu/trusty.task/ and ubuntu/xenial.task/ directories.

tasks/ubuntu/trusty.task/os_boot.erb
caseywilliams: Oh, I think I get it -- did these templates change for bionic (which is now default) and now the trusty task needs its own copies?

speedofdark:
Exactly.

tasks/ubuntu/bionic.task/metadata.yaml
```diff
@@ -0,0 +1,6 @@
+---
+
+os_version: bionic
+label: Ubuntu Bionic 18.04
+description: Ubuntu Bionic Installer
+base: ubuntu
```
smcclellan: It looks like `ubuntu/bionic.task` is missing the README (should be same as `ubuntu.task/README.md` in this case).

smcclellan:
This looks good once the README file is added. I'll also investigate better ways to do the networking work-around in the meantime.

smcclellan:
One other bit, this commit message should include which ISO works (since the live server doesn't work). It may not hurt even to add the full URL to the commit description.

tasks/ubuntu.task/os_boot.erb
```diff
@@ -4,6 +4,11 @@ exec >> /var/log/razor.log 2>&1
 
 <%= render_template("set_hostname") %>
 
+# ubuntu 18.04 networking doesn't wake up right away, so...
```
smcclellan: After more research, this still looks like the most straightforward solution. I do think it should move above the `set_hostname` template above, since the template reaches out to Razor server if something fails.

speedofdark: Done.

PR #430, opened 2019-07-09
Allow access to structured facts and metadata

smcclellan:
Modern versions of facter produce structured facts, or facts in a format
like hashes and arrays. If we are to utilize these facts in any way,
Razor's tagging system needs to be updated to allow access into those
facts. Since facts and metadata are similarly structured, the same
accessing method is added for both.

The `"fact"` and `"metadata"` matchers now use a dot-syntax for
scoping into hashes and arrays. Hashes will find the value from the
corresponding key, while arrays will look up the index. For example:

`["fact", "a.b"]` with a fact of `{"a" => {"b" => 1} }` will return
`1`.
`["fact", "a.1"]` with a fact of `{"a" => [1, 2] }` will return `2`.
`["fact", "a.length"]` with a fact of `{"a" => [1, 2, 3] }` will return
`3`.

If the scope does not exist, the `"fact"` and `"metadata"` matchers
will still use the third argument, the default, as prior to this change.

Fixes https://tickets.puppetlabs.com/browse/RAZOR-1110

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,7 +34,7 @@ gem "unix-crypt", "~> 1.1.1"
 
 
 group :doc do
-  gem 'yard', '~> 0.9.11'
+  gem 'yard', '>= 0.9.20'
   # kramdown 1.15.0 requires ruby >= 2.0
   gem 'kramdown', '~> 1.14.0'
 end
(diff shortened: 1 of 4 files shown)


smcclellan:
Follow-up: https://github.com/puppetlabs/razor-server/pull/432

PR #431, opened 2019-07-10
(maint) Update Travis' bundler version

smcclellan:


smcclellan:
Moving back instead of forward: https://github.com/puppetlabs/razor-server/pull/432

PR #436, opened 2019-09-17
Modifies Ubuntu 18.04 task.

rngkll:
Adds adds and modifies stages to get the ubuntu 18.04 nodes
to register.

### DIFF
--- a/tasks/ubuntu/bionic.task/os_complete.erb
+++ b/tasks/ubuntu/bionic.task/os_complete.erb
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+echo Razor policy successfully applied > /tmp/razor_complete.log
+echo Installed using <%= task.label %> - <%= task.description %> >> /tmp/razor_complete.log
+echo Repo: <%= repo_url %> >> /tmp/razor_complete.log
+echo Node: <%= node_url %> >> /tmp/razor_complete.log
+
+sed -i '/razor_postinstall/d' /etc/crontab
+
+curl -s <%= broker_install_url %> | bash
+if [ $? -ne 0 ]; then
+    echo "Broker run failed; see /var/log/razor.log for more details" >> /tmp/razor_complete.log
+else
+    echo Ran '<%= broker_install_url %>' >> /tmp/razor_complete.log
+fi
--- a/tasks/ubuntu/bionic.task/store_ip.erb
+++ b/tasks/ubuntu/bionic.task/store_ip.erb
@@ -1,0 +1,10 @@
+<%# -*- shell-script -*- %>
+# Get current IP
+node_ip=$(ip addr show | grep -A 2 BROADCAST | grep inet | awk '{ print $2 }' | awk -F'/' '{print $1}')
+echo IP is $node_ip
+
+# Send IP up
+curl -s <%= store_url("ip" => "$node_ip") %>
+
+# Send final state
+curl -s <%= stage_done_url("store_ip") %>
--- a/tasks/ubuntu/bionic.task/preseed.erb
+++ b/tasks/ubuntu/bionic.task/preseed.erb
@@ -1,0 +1,56 @@
+d-i console-setup/ask_detect boolean false
+d-i keyboard-configuration/layoutcode string us
+d-i netcfg/choose_interface select auto
+d-i netcfg/get_hostname string <%= node.metadata['hostname'] || node.hostname %>
+d-i netcfg/get_domain string <%= (node.metadata['hostname'] || node.hostname).split('.').drop(1).join('.') %>
+d-i netcfg/no_default_route boolean true
+# This is introduced in Trusty.
+d-i live-installer/net-image string <%= repo_url("install/filesystem.squashfs") %>
+d-i mirror/protocol string <%= repo_uri.scheme %>
+d-i mirror/country string manual
+d-i mirror/http/hostname string <%= "#{repo_uri.host}:#{repo_uri.port}" %>
+d-i mirror/http/directory string <%= repo_uri.path %>
+d-i clock-setup/utc boolean true
+d-i time/zone string <%= node.metadata['timezone'] || 'US/Pacific' %>
+d-i clock-setup/ntp boolean true
+d-i clock-setup/ntp-server string ntp.ubuntu.com
+d-i partman-auto/disk string /dev/sda
+d-i partman-auto/method string lvm
+d-i partman-lvm/device_remove_lvm boolean true
+d-i partman-md/device_remove_md boolean true
+d-i partman-auto-lvm/guided_size string max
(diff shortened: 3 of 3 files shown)


tasks/ubuntu/bionic.task/metadata.yaml
```diff
@@ -4,3 +4,6 @@ os_version: bionic
 label: Ubuntu Bionic 18.04
 description: Ubuntu Bionic Installer
 base: ubuntu
+boot_sequence:
```
smcclellan: This change shouldn't be needed. The sequence is inherited from `ubuntu.task/metadata.yaml`, which is identical to this.

tasks/ubuntu/bionic.task/os_boot.erb
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
smcclellan: Does this file differ from `ubuntu.task/os_boot.erb`? This file should also be inherited, rather than overwritten, if possible.

tasks/ubuntu/bionic.task/os_complete.erb
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+echo Razor policy successfully applied > /tmp/razor_complete.log
+echo Installed using <%= task.label %> - <%= task.description %> >> /tmp/razor_complete.log
+echo Repo: <%= repo_url %> >> /tmp/razor_complete.log
+echo Node: <%= node_url %> >> /tmp/razor_complete.log
+
+sed -i '/razor_postinstall/d' /etc/crontab
```
smcclellan: This looks like the only line that has changed so far. Was the other method of stopping the postinstall script not working?

tasks/ubuntu/bionic.task/store_ip.erb
```diff
@@ -0,0 +1,12 @@
...
+node_ip=$(ip addr show | grep -A 2 BROADCAST | grep inet | awk '{ print $2 }' | awk -F'/' '{print $1}')
+echo IP is $node_ip
+
+echo "$node_ip" > /tmp/store_ip.txt
+
+# Send IP up
+curl -s <%= store_url("ip" => "$node_ip") %>
+
+# Send final state
+curl -s <%= stage_done_url("finished") %>
```
smcclellan: This looks like a duplicate of `ubuntu.task/os_boot.erb`. Why move it into the storing of IP?

tasks/ubuntu/bionic.task/store_ip.erb
```diff
@@ -0,0 +1,12 @@
+<%# -*- shell-script -*- %>
+# Get current IP
+node_ip=$(ip addr show | grep -A 2 BROADCAST | grep inet | awk '{ print $2 }' | awk -F'/' '{print $1}')
+echo IP is $node_ip
+
+echo "$node_ip" > /tmp/store_ip.txt
```
smcclellan: Why put this on disk?

smcclellan:
Thank you for the pull request! I've added some comments to get this in shape. Overall it is more likely to be accepted if the commit description can call out specifics on what is changing, and why.

tasks/ubuntu/bionic.task/os_boot.erb
```diff
@@ -0,0 +1,31 @@
...
+apt-get -y dist-upgrade
+[ "$?" -eq 0 ] || curl -s <%= log_url("apt_dist_upgrade_fail", :error) %>
+
+
+<%= render_template("store_ip") %>
+
+<%= render_template("os_complete") %>
+
+# Send final state
+curl -s <%= stage_done_url("os_boot") %>
```
rngkll: the one inherited is sending the finished message not specifying the state.

is "finished" ok in this case?

tasks/ubuntu/bionic.task/os_complete.erb
rngkll: post install script was never called because ubuntu 18.04 uses systemd instead of init.

adding a line to the crontab works with systemd.

tasks/ubuntu/bionic.task/store_ip.erb
rngkll: this was for checking the correct IP only, I have to remove this.

tasks/ubuntu/bionic.task/store_ip.erb
rngkll: this stage the last one, by adding this I got all the stages to run and the nodes get registered.

tasks/ubuntu/bionic.task/metadata.yaml
rngkll: I'll remove this one.

Thanks,

rngkll:
@smcclellan I did some changes based on  your comments, let me know if this helps.

Thanks,

tasks/ubuntu/bionic.task/os_complete.erb
smcclellan: Ah excellent, that makes sense. Is there a reason this wouldn't work for other Ubuntu versions too? This might be a more portable way to be standard across versions.

tasks/ubuntu/bionic.task/store_ip.erb
smcclellan: I would expect a `"finished"` log message to be reported to the server already, via this in `os_boot.erb` (which renders this file):
```
<%= render_template("store_ip") %>

<%= render_template("os_complete") %>

# Send final state
curl -s <%= stage_done_url("finished") %>
```
Is that last `"finished"` message not showing up?

tasks/ubuntu/bionic.task/os_boot.erb
smcclellan: Specifying the state, you mean stating that the task is finished?

tasks/ubuntu/bionic.task/os_complete.erb
rngkll: I haven't tested this with other than ubuntu 18.04, this might work for newer version but not sure for older.

tasks/ubuntu/bionic.task/store_ip.erb
rngkll: my mistake, i removed this.

ashish1099:
any update on this one ?

rngkll:
> any update on this one ?

I fairly don't remember if I signed or not, I think I did because I have no problem with signing the CLA.

I never did the test with other debian based or debian, but it should work, in some time, I'll use razor server again and will like to test them.

https://cla.puppet.com/ is not available, can someone point me to the new link.

smcclellan:
FYI I'm no longer involved with Razor, and not sure entirely who the POC for the product is now. @McdonaldSeanp may know perhaps?

PR #443, opened 2020-07-02
(SEC-166) update jbdc postgres, rack

steveax:


Gemfile
```diff
@@ -21,12 +21,12 @@ ruby '2.3.1', :engine => 'jruby', :engine_version => '9.1.5.0'
 gem 'torquebox', '~> 3.2.0'
 gem 'sinatra'
 gem 'sequel'
-gem 'jdbc-postgres', '= 42.2.6'
+gem 'jdbc-postgres', '= 42.2.14'
```
donoghuc, line 24: this is equivalent to `gem 'jdbc-postgres', '42.2.14'`

donoghuc (approved):
I traced this through and pinning the Gemfile is the correct action (though i'm not convinced the gemfile.lock does anything meaningful). Essentially in vanagon we bundle install, then ship artifcacts with the bundler structure which somehow appears to be included in torquebox. This will update the right packages but the packaging for this does not follow patterns for existing ruby/jruby projects that i am aware of. :shrug:

